PR #7, opened 2018-11-22
Add an iterative variant of the run method to mcmc

joezuntz:
This change adds an iterative version of the mcmc.run method that uses the python "yield" keyword to send each sample to the caller during the sampling process, rather than just at the end.

This can help if people (like me!) using the code as a library want to handle I/O themselves but also want to progressively output results, rather than just at the end.

To avoid copying code I have split the "run" method into three pieces, and now call each of those from both the new version of "run" and the new "iterate" method.

The only possible point of confusion is that I have set the weight of the returned sample to 1.0 in this case, because this method will return the same sample several times if the proposal is not accepted each time.  This is only done on a copy of the sample.

### DIFF
--- a/cobaya/samplers/mcmc/mcmc.py
+++ b/cobaya/samplers/mcmc/mcmc.py
@@ -308,14 +308,12 @@ class mcmc(Sampler):
         assert not np.any(np.isnan(covmat))
         return covmat
 
-    def run(self):
-        """
-        Runs the sampler.
-        """
+    def _start_run(self):
         # Get first point, to be discarded -- not possible to determine its weight
         # Still, we need to compute derived parameters, since, as the proposal "blocked",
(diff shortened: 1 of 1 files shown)


JesusTorrado:
Hi Joe,

Thanks, it looks good! I'll look at merging next week (busy with teaching for a few days straight)

JesusTorrado:
Hi Joe,

Good! In that case, I'll close this for now, and we can re-open it if you find that there is something that cannot be done that way.

Cheers!

JesusTorrado:
@joezuntz I have added in the last commit (1f9a8fe) an option to get the last chain points added more easily, and added [an example to the documentation](https://cobaya.readthedocs.io/en/latest/sampler_mcmc.html#callback-functions). I am in the process of doing the same for PolyChord, in case you want to use it.

PR #21, opened 2019-01-19
Updates for new CAMB 1.0

cmbant:
There are some problems here with installing polychord (does it rely on utils.F90 from the old CAMB?), and with getting DH and other derived parameters ("Could not resolve arguments for derived parameters ['DHBBN']") - do you know what might be causing that as not so obvious?

Main changes are: pycamb folder removed, now root. CAMBparams now has ombh2, omk etc as base parameters (omegam etc are read-only properties so no real change) - but Omega_de (mapped from omegal) is now only in the CAMBdata results object.  Also CAMBparams is now effectively a read-only input that is not changed by the calculation, so I've changed to getting derived attributes from the results.Params object (which has any additional things calculated).

"forutils" is now a submodule of camb - this makes installing CAMB from a github tar.gz much messier, since github doesnt't include the submodule.  I've made the camb setup.py pull forutils if needed if it is not already installed, and this seems to work. However ultimately would probably be much better to do recursive clone with git (as in the test scripts).

Work still to be done to access new thinks like setting up SourceWIndows array and specifying classes for dark-energy models etc more efficiently..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: python
-dist: trusty
+dist: xenial
 sudo: required
 
 # (Pre)Installation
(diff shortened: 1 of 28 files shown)


cmbant:
This pull request is now superseded by the one including clik update.

PR #22, opened 2019-02-26
Updates for CAMB 1.0 and latest Planck 2015 clik code supporting py3 etc.

cmbant:


JesusTorrado:
Merged into 1.1 devel branch (we can already tell users to use it). Still reviewing changes.

Is this retro-compabitle with CAMB < 1.0?

cmbant:
It's not backwards compatible, I don't suggest we try to be (hence the 1.0 name for the new camb version, so will try to maintain consistency from now)

PR #53, opened 2019-11-08
External modules

cmbant:
This adds support for loading likelihood modules from an external location, either on the python path or at a specified folder location. There's a brief doc in likelihoods.rst (would be nice if we could turn on readthedocs for at least one test branch):

Implementing your own *external* likelihood class
--------------------------------------------------

Instead of including the likelihood within the standard Cobaya likelihood modules, you may wish to make an external
package that can be redistributed easily. To do this you make a package containing a class defined exactly the same way
as for internal likelihoods above (inheriting from :class:`Likelihood` as documented below). For example if you have a
package called *mycodes*, containing a likelihood class called MyLike in *mycodes.mylikes*, when running Cobaya you can
use the input

     likelihood:
       mycodes.mylikes.MyLike:
         [option 1]: [value 1]
         [...]

This is assuming that *mycodes* is on your Python path (e.g. it is an installed package).
You can also specify an explicit path for the module, e.g.

     likelihood:
       mycodes.mylike.MyLike:
         python_path: /path/to/mycodes_dir
         [option 1]: [value 1]
         [...]

For an example class implementation, check out :doc:`cosmo_external_likelihood_class`.
There is also a simple [external likelihood package](https://github.com/CobayaSampler/example_external_likelihood)
and a real-word cosmology example in the [sample external CMB likelihood](https://github.com/CobayaSampler/planck_lensing_external).

PR #60, opened 2019-11-20
add ability to define ._defaults class attribute for Likelihood subclass

timothydmorton:
Haven't added any tests yet, but this works for the simple example I've been playing with.  I figured I'd put this in `Likelihood` rather than in `HasDefaults` just because I didn't know what else uses `HasDefaults`.  You can define a `._defaults` dictionary as a class attribute and a YAML file will not be required.  Basing off `external_modules` branch because that's where I'm working.

### DIFF
--- a/cobaya/likelihood.py
+++ b/cobaya/likelihood.py
@@ -87,8 +87,12 @@ class Likelihood(HasLogger, HasDefaults):
             logging.info('No YAML defaults available; relying on ._defaults class attribute if provided.')
             defaults = odict([('likelihood', odict([('__self__', {})]))])
 
-        class_defaults = odict(cls._defaults) if cls._defaults is not None else {}
-        defaults['likelihood']['__self__'].update(class_defaults)
+        # Specifically for externally defined likelihoods?
+        if 'likelihood' in defaults:
+            if '__self__' in defaults['likelihood']:
+                class_defaults = odict(cls._defaults) if cls._defaults is not None else {}
+                defaults['likelihood']['__self__'].update(class_defaults)
+
         return defaults
 
     # Optional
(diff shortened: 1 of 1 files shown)


timothydmorton:
PS I wasn't sure how to get a more useful logging message, since `cls.__name__` just showed up as `Likelihood` rather than the subclass I was using in the example.

timothydmorton:
crossreference #53

cmbant:
If the "devel" branch the CobayaComponent base class already has a class_options dictionary. Allowing no yaml file should probably also be done generally in CobayaComponent (or indeed HasDefaults).
(note super() is also not py27 compatible, though hopefully soon we can drop support for that).

timothydmorton:
does 'devel' include the 'external_modules' updates?  I've been working on the latter, but maybe I should be on the former?

timothydmorton:
ah, sorry, hadn't seen #57

timothydmorton:
@cmbant @JesusTorrado: just FYI, I'm working on allowing for no YAML (unless you tell me you're working on it).  I'm now based off of devel; will make PR when ready.

cmbant:
https://github.com/CobayaSampler/cobaya/pull/61/commits/490295711190a01be3c48d4ac1e2b1dc6d52f9a7 implements something like this. The main differences are that it uses cls.__dict__['class_options'] or .yaml for each class before merging with inherited (cls.class_options will take inherited class options if none defined in this class, cls.__dict.get('class_options') will return None if in this class). I have also changed the .yaml defaults format to all be class-level variables, i.e. no `likelihood: __self__`. This simplify several things, and allows likelihoods to inherit from theories. Does this achieve what you wanted? All help testing appreciated - some of these things are large changes.

PR #61, opened 2019-11-20
Devel

cmbant:
This is the latest development branch, including others. Includes external likelihoods, multiple theories, complex dependencies, class/camb fixes optimizations, etc.


cmbant:
Support for multiple theories with complex dependencies is now pretty much done. Note the yaml defaults format has changed: no `likelihood: __self__` embedding any more, and class_options and yaml settings are now inherited by subclassed likelihoods. Needs lots of help testing!

This test shows how to set up multiple theories in a realistic case - i.e. making the BBN consistency model a separate theory code: https://github.com/CobayaSampler/cobaya/blob/devel/tests/test_cosmo_multi_theory.py
In principle CCL+CAMB integration should now be possible, but I have not tried that specific case.

cmbant:
We now have both a component "version" field, which is a version string, and a get_version method which in general returns a dictionary - this could be confusing? (I went for a dictionary since for camb etc it would be nice to be able to return versions of all extra components to document their versions (halofit, extra custom classes etc, and likelihoods might have both code and data versions).

Issue #64, opened 2019-12-02
Avoid trying to import MPI for installing


timothydmorton:
In the process of ensuring that it only runs on one MPI instance, the install script currently attempts to `from mpi4py import MPI`.  If mpi4py is not available, this is all well and good, as the try/except works as intended.  However, if trying to run `cobaya-install` from a system that has mpi4py but is disallowed from using it (e.g., the head nodes at NERSC), then `cobaya-install` fails because `from mpi4py import MPI` throws a fatal error that aborts the script, rather than simply raising an exception.  

It appears that moving `from cobaya.mpi import am_single_or_primary_process` further inside the script, protected by a '--no-mpi' argument check (e.g.) should work.  Happy to make a PR for this.

timothydmorton:
Should I base this PR off master or the 'temp' branch?

timothydmorton:
I see this issue has its tentacles in multiple places (i.e., wherever `am_single_or_primary_process` is called); I'm looking into it.

cmbant:
Thanks, merged in temp. I'd use devel for PRs unless specific to something being developed in a temporary side branch.

PR #65, opened 2019-12-02
Protect attempted MPI import

timothydmorton:
On a system where mpi4py is available but disallowed, the command 'from mpi4py import MPI' can cause an ungraceful script abort (e.g., on a head node at NERSC). This circumvents this for the cobaya-install script such that it can be run on a head node at NERSC.  I have tested that cobaya-install now works on head node at NERSC with this update.

As this issue also shows up in setting up the logger, the logger_setup call in the install script needed to be moved until after the arguments are parsed, to allow it to be protected.

### DIFF
--- a/cobaya/install.py
+++ b/cobaya/install.py
@@ -239,39 +239,48 @@ def check_gcc_version(min_version="6.4", error_returns=None):
 # Command-line script ####################################################################
 
 def install_script():
+    warn_deprecation()
+
+    # Parse arguments
+    import argparse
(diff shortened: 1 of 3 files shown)


timothydmorton:
Addressing #64

JesusTorrado:
Thanks, @timothydmorton !

Antony is in control of that branch, so let him merge it.

cmbant:
I also pushed on devel, so it doesn't get lost in squash merge from temp.

PR #67, opened 2019-12-07
Split camb into two Theory classes

cmbant:
Allows semi-slow parameters for the initial power spectrum or nonlinear model, generalizing cosmomc's approach to also work when non-linear corrections are needed (by computing time transfer functions, and recalculating nonlinear corrections, bessel integrals and lensed CL when the semi-slow parameters change). This should make cobaya faster than cosmomc, since compute saving should be significant. 

To do this the camb class now dynamically generates a helper transfer function Theory class on which it depends, and assigns initial power/nonlinear parameters to itself and other parameters to the transfer function class.

Not sure how this works with the updated yaml, resuming etc, could do with some help on that. (is resuming/checkpointing with camb in any of the tests?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@
 - changed use_planck_names to more general use_renames etc.
 - Greatly reduced deepcopy time overhead in loglike evaluation, faster uniform prior and sample collection
 - Global stop_at_error option to stop at error in any component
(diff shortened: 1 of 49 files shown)


cmbant:
I did some tests on resuming and it actually works. The "camb" module appears in the updated yaml with only initial power/non-linear parameters and other camb transfer parameters do not appear. But it seems to work since camb transfers supports agnostic, so still gets the correct parameters on running. Not sure if necessarily ideal though, or implications elsewhere.

PR #69, opened 2019-12-11
Better fix for MPI crashing 

timothydmorton:
As mentioned in #64 and partially addressed in #65, using cobaya on the head node of a cluster where MPI is disallowed causes python to crash ungracefully.  Since fixing #64, I've run into this issue again trying to use cobaya interactively.  I think this solution is better, as it allows an interactive user to do the following:
```python
from cobaya.mpi import disable_mpi
disable_mpi()
```
and then go on with life.  Followed by (if desired, in some crazy world?)
```python
from cobaya.mpi import enable_mpi
enable_mpi()
```
to restore the globals to their defaults.

### DIFF
--- a/cobaya/install.py
+++ b/cobaya/install.py
@@ -273,11 +273,15 @@ def install_script():
                             help="Install data of the modules.", dest=_code)
     arguments = parser.parse_args()
 
+    if arguments.no_mpi:
+        from cobaya.mpi import disable_mpi
+        disable_mpi()
+
     from cobaya.mpi import am_single_or_primary_process
-    is_primary_process = am_single_or_primary_process(no_mpi=arguments.no_mpi)
+    is_primary_process = am_single_or_primary_process()
 
     # Configure the logger ASAP
-    logger_setup(no_mpi=arguments.no_mpi)
+    logger_setup()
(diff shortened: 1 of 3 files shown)


cmbant:
In https://github.com/CobayaSampler/cobaya/pull/67 I had added a "disabled" option to mpi, e.g.

```
from cobaya import mpi
mpi.disabled =False
```
with optional env variable to set automatically - does that also work for you?
Of course the nicest thing would be to programmatically tell if mpi will work, but I can't find a way that won't hang on some systems.

cmbant:
PS, couldn't no-mpi be made default on cobaya-install?

JesusTorrado:
Maybe it does make sense to make `install` mpi-UNaware and dropping all mpi checks. Ok from me.

timothydmorton:
Yeah, that all sounds fine to me!  Sorry for the duplicate PR; I missed the update in #67.  @JesusTorrado I agree that unless there's a good reason to be otherwise, `cobaya-install` should always just disable MPI checks, with no argument necessary.

cmbant:
I've disabled mpi in install in latest https://github.com/CobayaSampler/cobaya/pull/67

Issue #72, opened 2019-12-31
API change request to expose model likelihoods


msyriac:
When a new model is built with
```
model = get_model(info)
```

it is useful to have the underlying likelihoods exposed to the user. An example where I needed this was to generate mock BAO likelihood data for pipeline verification. The BAO likelihood has a useful `theory_fun` (e.g. to get DM_over_rs) which I didn't want to duplicate. It was not obvious through a cursory tab completion search in iPython on `model.likelihood` where I could access the underlying likelihood objects, and I had to dig through the code to find that it could be accessed through e.g:

```
thfn = model.likelihood._likelihoods['bao.sdss_dr12_consensus_bao'].theory_fun                                                                                                                
```

Attributes with an underscore in front of them are not discoverable through tab completion (unless an underscore is explicitly prefixed), so I recommend that the underscore be removed.

If there was a better way to access that underlying theory function, do let me know.

msyriac:
UPDATE:
It appears more changes are needed to expose those functions in a usable way. (Either that, or I'm taking unnecessarily convoluted steps).

If the above theory function is used, one gets:
```
Traceback (most recent call last):
  File "gen_bao.py", line 38, in <module>
    print(thfn(z,obs))
  File "/home/r/rbond/msyriac/.local/lib/python3.6/site-packages/cobaya-2.0.3-py3.6.egg/cobaya/likelihoods/_base_classes/_bao_prototype.py", line 284, in theory_fun
    val = (1 + z) * self.theory.get_angular_diameter_distance(z) / self.rs()
  File "/home/r/rbond/msyriac/.local/lib/python3.6/site-packages/cobaya-2.0.3-py3.6.egg/cobaya/theories/camb/camb.py", line 627, in get_angular_diameter_distance
    return self._get_z_dependent("angular_diameter_distance", z)
  File "/home/r/rbond/msyriac/.local/lib/python3.6/site-packages/cobaya-2.0.3-py3.6.egg/cobaya/theories/camb/camb.py", line 616, in _get_z_dependent
    return np.array(deepcopy(self.current_state()[quantity]))[i_kwarg_z]
KeyError: 'angular_diameter_distance'
```

because the underlying theory calls have not been made. So to get it to work, I needed to do:
```
model = get_model(info)                                                                                                                                                                       
updated_info, products, sampler = run(info)                                                                                                                                                  
thfn = sampler.model.likelihood._likelihoods['bao.sdss_dr12_consensus_bao'].theory_fun
[...]

PR #73, opened 2019-12-31
Expose _likelihoods and add option for returning full sampler object from run.run()

msyriac:
This addresses issue #72. Note that I've worked off the `sources` branch of CobayaSampler/cobaya.

It changes the `_likelihoods` attribute of `LikelihoodCollection` to `likelihoods` to make it discoverable in tab completion. It also adds an option for cobaya.run.run() to return the full sampler object so that the specific theory functions from a likelihood can be explicitly accessed. e.g. for accessing the BAO likelihood theory functions:

```
model = get_model(info)                                                                                                                                                                       
updated_info, products, sampler = crun(info,return_sampler=True)                                                                                                                              
DM_over_rs = sampler.model.likelihood.likelihoods['bao.sdss_dr12_consensus_bao'].theory_fun(z=0.51,'DM_over_rs')
```

If there are better existing ways to achieve this, I'll close this PR.

### DIFF
--- a/cobaya/likelihood.py
+++ b/cobaya/likelihood.py
@@ -289,15 +289,15 @@ class LikelihoodCollection(HasLogger):
         else:
             self.theory = None
         # Initialize individual Likelihoods
-        self._likelihoods = odict()
+        self.likelihoods = odict()
         for name, info in info_likelihood.items():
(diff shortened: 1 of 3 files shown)


msyriac:
UPDATE: looking into branching this off the devel branch.

PR #84, opened 2020-04-09
model: add_manual_requiremnt, simple implementation

JesusTorrado:

### DIFF
--- a/cobaya/model.py
+++ b/cobaya/model.py
@@ -503,7 +503,7 @@ class Model(HasLogger):
 
         self._component_order = {c: components.index(c) for c in dependence_order}
 
-    def _set_dependencies_and_providers(self):
+    def _set_dependencies_and_providers(self, manual_requirements={}):
         requirements = []
         dependencies = {}
         self._needs = {}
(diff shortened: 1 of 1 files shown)


cobaya/model.py
```diff
@@ -521,7 +521,7 @@ def _tidy_requirements(_component, _require):
             else:
                 _require = dict.fromkeys(_require)
             for par in self.input_params:
-                if par in _require:
+                if par in _require and component is not None:
```
cmbant, line 524: should this be _component?

cobaya/model.py
```diff
@@ -503,15 +503,15 @@ def _set_component_order(self, components, dependencies):
 
         self._component_order = {c: components.index(c) for c in dependence_order}
 
-    def _set_dependencies_and_providers(self):
+    def _set_dependencies_and_providers(self, manual_requirements={}):
```
cmbant, line 506: {} to empty_dict to be immutable argument

cobaya/model.py
```diff
@@ -551,6 +551,12 @@ def _tidy_requirements(_component, _require):
 
             for k in can_provide + component.output_params + provide_params:
                 providers[k] = providers.get(k, []) + [component]
+        # Add requirements requested by hand
+        _manual = '_manual'
```
cmbant, line 555: Could just use None rather than '_manual'? Looks odd mixing strings and Components

PR #85, opened 2020-04-10
model: simple fix

JesusTorrado:
Basic implementation. Does not change any actual behaviour, but allows to use the "requires" option of a theory/likelihood input to indicate compoments.


cobaya/theory.py
```diff
@@ -75,21 +75,64 @@ def get_requirements(self):
...
+        reqs = self.get_requirements()
+        if not isinstance(reqs, Mapping):
+            reqs = dict.fromkeys(reqs)
+        for req, opt in reqs.items():
+            if opt is None:
+                if getattr(self, "provider", None) is not None:
+                    if req in self.provider.requirement_providers:
+                        this_provider = self.provider.requirement_providers[req]
+                        if req in this_provider.get_can_provide_params():
+                            self._required_params.add(req)
```
cmbant, line 99: This peeking inside the implementation of Provider (an indirectly using properties of other theories) is a bit odd, and not really in the spirit of class separation.

[without this, even the existence of self.provider is an implementation choice rather than enforced - just the default initialize_with_params() implementation chooses to store the Provider instance into self.provider as a default convention]

JesusTorrado:
> Btw, we should perhaps allow all requirements to also be arrays of Requirement or (key, option) tuples (this is already done internally for combining requirements, but currently no component can individually return two requests with the same key but different options because dicts have unique keys).

Added to the TODO list.

cobaya/theory.py
JesusTorrado: This is not done any more (this method has been removed), since it is actually not necessary at least for the cases we work with right now to distinguish between parameters and other requirements.

cobaya/model.py
```diff
@@ -537,96 +535,99 @@ def _tidy_requirements(_require, _component=None):
...
+                        self._must_provide[supplier] + must_provide[supplier]
+                    if requirement not in declared_requirements_for_this_supplier:
+                        must_provide[supplier] += [requirement]
+                    dependencies[component] = \
+                        dependencies.get(component, set()) | {supplier}
+            # tell each component what it must provide, and collect the
+            # conditional requirements for those requests
+            there_are_more_requirements = False
+            for component, requires in requirements.items():
+                # empty the list of requirements, since they have already been assignes,
```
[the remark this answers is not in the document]
JesusTorrado: Corrected, thanks!

cobaya/model.py
```diff
@@ -537,96 +535,99 @@ def _tidy_requirements(_require, _component=None):
...
             provide_params = [p for p in component.get_can_provide_params() if
-                              p not in self.output_params and p not in
-                              str_to_list(getattr(component, _requires, []))]
-
+                              p not in self.output_params]
+            # Corner case: some components can either take some parameters as input OR
+            # provide their own calculation of them. Pop those if required as input.
+            for p in provide_params.copy():  # iterating over copy
+                if p in component.get_required_params():
+                    provide_params.remove(p)
```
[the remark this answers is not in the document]
JesusTorrado: I think so, but clearer (at least for me).

cobaya/model.py
```diff
@@ -235,9 +235,9 @@ def logps(self, input_params, return_derived=True, cached=True, make_finite=Fals
         self.log.debug("Got input parameters: %r", input_params)
         n_theory = len(self.theory)
         loglikes = np.empty(len(self.likelihood))
-        for (component, index), dependence in zip(self._component_order.items(),
-                                                  self._ordered_param_dependence):
-            depend_list = [input_params[p] for p in dependence]
+        for component, index in self._component_order.items():
+            depend_list = \
+                [input_params[p] for p in self._params_of_dependencies[component]]
```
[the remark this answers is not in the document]
JesusTorrado: Restored old behaviour in 25c38f8

cobaya/theory.py
```diff
@@ -204,11 +213,13 @@ def check_cache_and_compute(self, dependency_params=None, want_derived=False,
         (retrieved using get_current_derived()).
         """
         self.log.debug("Got parameters %r", params_values_dict)
-
-        for set_param in getattr(self, _requires, []):
+        params_values_dict = params_values_dict.copy()
+        for set_param in self.get_requirements():
```
cmbant, line 217: Perhaps we could precompute a list of which things in get_requirements are not in input_params and look like parameters, so the copy and loop can still be avoided in most cases?

cobaya/model.py
```diff
@@ -727,7 +731,12 @@ def _assign_params(self, info_likelihood, info_theory=None):
                     provide = str_to_list(getattr(component, _provides, []))
                     supports_params |= set(provide)
                 else:
-                    supports_params = set(component.get_requirements()).union(
+                    required_params = ensure_dict(component.get_requirements()).copy()
+                    # pop non-params; it's ok if some non-param goes through
+                    for p,v in required_params.copy().items():
+                        if v:  # not a param
+                            required_params.pop(p)
+                    supports_params = set(required_params).union(
```
cmbant, line 739:
```
                    required_params = set(p for p, v in
                                       ensure_dict(component.get_requirements()).items()
                                       if v is None)
```?

cobaya/theory.py
JesusTorrado: Done in 39bcf36 as a property.

PR #88, opened 2020-04-16
add get_cobaya_class load path

cmbant:
Thinking about how e.g. polychord could be made entirely external, and how other external general packages like CCL could support a cobaya class, it may be useful for external python packages to be able to provide a cobaya class without depending on Cobaya or loading it by default. So this adds a `get_cobaya_class() `function support to `get_class`, which could dynamically load the relevant class from the package `main __init_`_ file (i.e. if "polychord" external python package defined this in its `__init__`, it could then be used with exactly the same syntax as now as an internal package).

[Could also be used to e.g. rename internal classes with PEP8 formatting, e.g. MCMC rather than mcmc, loading from `mcmc.__init__ `and overriding `MCMC.get_qualified_class_name` to return "mcmc"; but there are likely complication in other places where the class name is used, e.g. loading the bib/yaml]

### DIFF
--- a/cobaya/conventions.py
+++ b/cobaya/conventions.py
@@ -11,7 +11,7 @@ from types import MappingProxyType
 
 # Package name (for importlib)
 # (apparently __package__ is only defined if you import something locally.
-_package = __name__.rpartition('.')[0]
+_cobaya_package = __name__.rpartition('.')[0]
 
 # an immutable empty dict (e.g. for argument defaults)
 empty_dict = MappingProxyType({})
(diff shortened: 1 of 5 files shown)


JesusTorrado:
I don't see anything wrong with it, so go ahead.

For PolyChord specifically, being a "core" component, I'd rather maintain the interface myself.

PEP8 is a style guide, not a mandatory set of rules. If there are strong reasons to ignore some of its guidelines (e.g. naming conventions in a particular field) it is more intuitive to keep it so in the code, I would say. Not that we are using those conventions everywhere ("mcmc" is a good counterexample) in any case (my bad).

PR #92, opened 2020-05-12
A few install improvements

JesusTorrado:
- Swaps `wget` for `requests` (better maintained, apparently less prone to segfaults). Motivated by https://cosmocoffee.info/viewtopic.php?f=11&t=3344
- Files now downloaded to tmp folder before decompression, and automatically cleaned up
- Some var names tidies

@cmbant Could you please test it on Windows? just ``cobaya-install cosmo``

----

TO-DO list for the remaining of the PR

- [x] modify `is_installed` to have this dual behaviour, and adapt initialisation
  - [x] CAMB
  - [x] CLASS
  - [x] clik
  - [x] polychord (also, better error msgs)
- [x] add `--skip-global` to install -- tries an `is_installed` with global first
- [x] add `is_installed` to the tests, and cmdarg to skip global
- [x] simplify code:
  - [x] have initialize raise NotInstalledError (inheriting from LoggedError)
  - [x] remove check_installed where possible
- [x] add a non-global test if explicit path given
- [x] auto-skip incompatible in Windows
  - [x] classy
  - [x] clik
  - [x] polychord
- [x] document new behaviour and cmdarg
- [x] update changelog


cmbant:
Tried for camb and BAO and seems to work fine on Windows

cmbant:
(cosmo fails because classy [and clik[ fails on windows)

JesusTorrado:
Thanks for testing it!

I have also changed --just-check to --test in cobaya-install, for consistency with cobaya.run.

Another long-term install/test TODO on my list that could go in this PR would be an option having pytest **not** failing when trying to test something not installed, if requested to behave like that (we'd create a pytest fixture --only-installed or sth like that). This way, we wouldn't have to set `COBAYA_TEST_SKIP` in many cases (e.g. in Travis), just do `pytest --only-installed`

The hard part there is deciding at run-time which tests can fail. One way would be manually (or somehow programmatically) *marking* tests according to what they run.

An easiest one would be having all tests call `install` with `test=True` somewhere in their code, so that installation is tested but nothing is done for uninstalled modules. `install` would raise a custom `NotInstalledError` exception, which would be caught [like this](https://docs.pytest.org/en/latest/skipping.html#raises-parameter), and the test would be allowed to fail. This one is minimally invasive and easy to add to existing and new tests.

What do you think? Do you think is it worth it?

cmbant:
I guess people might want to test with locally installed camb etc versions, which are not in modules, which would be harder to make work with install(tesd=True) - though actually I'm not sure if that currently works or not? Tests on Windows could also skip incompatible.

JesusTorrado:
TO-DO list for the remaining of the PR

- [ ] make `path=global` get to `is_installed`
- [ ] modify `is_installed` to have this dual behaviour, and adapt initialisation
  - [ ] CAMB
  - [ ] CLASS
  - [ ] clik
  - [ ] polychord (also, better error msgs)
- [ ] add cmdarg to the test
- [ ] add `is_installed` to the tests

JesusTorrado:
Done. Now one can use `path=global` and it will fail if there is no global python module for the external dependency.

Also, now all tests have some installation check somewhere. Adding the pytest arg `--skip-not-installed`, tests with unsatisfied dependencies will be allowed to fail. So one does not need to use `COBAYA_TEST_SKIP` so often.

One question re planck: the `has_clik` think at initialisation is just to try a global import before a local install check, right? Or is there anything I am missing?

The only thing left in this PR, aside from docs, would be restoring download progress bars, which now need external dependencies, because Requests does not have their own. But we can live without them for now.

JesusTorrado:
I forgot: you can also do `cobaya-install --skip-global` and e.g. CAMB will not be installed if a global installation test passes (in this case, just checkin that `import camb`, without modifying the pythonpath, does not raise an exception).

cmbant:
Sounds good. Looks like has_clik is to make sure it doesn't import global if explicit path given, but otherwise try global before looking for separate installation in packages folder.

JesusTorrado:
Thanks! I should actually do that test for all of them. Added to the to-do list, together with some code simplifications to remove the need to add anything special to the tests.

JesusTorrado:
Done here!

@cmbant could you please check that the OS-compatibility auto-skip works on Windows? (just do `cobaya-install cosmo`, without `--skip` or `COBAYA_INSTALL_SKIP`)

cmbant:
Seems to work OK! The "ERROR* lines ("The given path does not exist", "Installation test failed"..) are probably a bit confusing to the uninitiated, as not detecting something already being installed is not really an error if nothing is installed (esp. if the packages folder doesn't already exist) - perhaps change message/hide from the install routine?

JesusTorrado:
> Seems to work OK! The "ERROR* lines ("The given path does not exist", "Installation test failed"..) are probably a bit confusing to the uninitiated, as not detecting something already being installed is not really an error if nothing is installed (esp. if the packages folder doesn't already exist) - perhaps change message/hide from the install routine?

Done. These messages are hidden unless a --debug cmd arg is passed (suggested if errors).

> Might also be useful to explicitly give the "global config file" path at the end of the install, so people know where it is?

Done.

Ok, to merge?

cmbant:
Thanks, looks good to me (pushed trivial tidies).

PR #95, opened 2020-05-22
add sigma_R to classy (WIP)

timothydmorton:
Wanted to open this to get feedback; haven't checked if this matches the expected format of `sigma_R` from CAMB.  I also tried to mimic how this was done on the CAMB side, and I don't understand much of it, but this does seem to successfully give back the sigma computed by classy.  Let me know if you have suggestions!

### DIFF
--- a/cobaya/theories/classy/classy.py
+++ b/cobaya/theories/classy/classy.py
@@ -218,6 +218,10 @@ class classy(BoltzmannBase):
         # Derived parameters that may not have been requested, but will be necessary later
         self.derived_extra = []
 
+    # def add_to_redshifts(self, z):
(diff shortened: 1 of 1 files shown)


cmbant:
Would be nice to add unit test analogue of test_CAMB_sigma_R to check it works with the same syntax (and gives same result!)

timothydmorton:
Agreed!  Will do so at my earliest convenience.

cmbant:
@timothydmorton ping for test?

JesusTorrado:
Done in https://github.com/CobayaSampler/cobaya/pull/222, with a slightly different syntax (using `arg_array` to get parameter combinations).

Issue #96, opened 2020-05-25
Issue in number of chain files


surgithub:
Hi, I am new to cobaya.

The command
`mpirun -np 4 cobaya-run planck1.yaml -p cobaya_packages`
initiates 4 instances of mpi, and runs fine. But in the output folder, I see only one chain file. 

Thanks.

cmbant:
Could you try with "mpi_warnings" branch, where I tried to add some error checks, and see if that give a more helpful error ?

@JesusTorrado, this has come up a few times in multiple guises. Do you think the check at https://github.com/CobayaSampler/cobaya/blob/mpi_warnings/cobaya/collection.py#L349 is safe (or better place to try to check exclusive write on each chain output)?

surgithub:
As mentioned earlier, it initiates 4 instances of mpi and runs fine without displaying any errors. I notice that all the accepted samples of four different samplings are written in one chain file. So I think the code needs some correction for writing the accepted samples in four different chain files. Thanks.

cmbant:
Did you install and run in the mpi_warnings branch?

surgithub:
Yes. I did that. Still the output is going to only one chain file. Following is the onscreen output. 

`mpirun -np 2   cobaya-run planck.yaml -p cobaya_packages -o trial/sk
--------------------------------------------------------------------------
[[43768,1],0]: A high-performance Open MPI point-to-point messaging module
was unable to find any relevant network interfaces:

Module: OpenFabrics (openib)
  Host: hn001

Another transport will be used instead, although this may result in
lower performance.

NOTE: You can disable this warning by setting the MCA parameter
btl_base_warn_component_unused to 0.
--------------------------------------------------------------------------
--------------------------------------------------------------------------
[[43771,1],0]: A high-performance Open MPI point-to-point messaging module
was unable to find any relevant network interfaces:

Module: OpenFabrics (openib)
  Host: hn001

Another transport will be used instead, although this may result in
lower performance.

NOTE: You can disable this warning by setting the MCA parameter
btl_base_warn_component_unused to 0.
--------------------------------------------------------------------------
[output_mpi] Output to be read-from/written-into folder 'trial', with prefix 'sk'
[output_mpi] Output to be read-from/written-into folder 'trial', with prefix 'sk'
[prior] *WARNING* External prior 'SZ' loaded. Mind that it might not be normalized!
[classy] Importing *local* classy from /home/hpadmin/cobaya_packages/code/classy
[...]

PR #99, opened 2020-05-28
add some warnings if attempting to produce same output without using MPI

cmbant:
Have heard non-mpi run problem from a few places now, should trap and give helpful error if possible. This PR may not be sufficient, c.f. https://github.com/CobayaSampler/cobaya/issues/96


cmbant:
Why reverting travis chages, seem redundant?

PR #100, opened 2020-06-01
add FIRAS units

cmbant:
Experiments like Planck are usually calibrated to FIRAS not an absolute temperature. This adds FIRAS units option to get C_l result in FIRAS muK^2, which is default for Planck (and hopefully making it easier to directly run varying T_CMB consistently).

### DIFF
--- a/cobaya/theories/_cosmo/boltzmannbase.py
+++ b/cobaya/theories/_cosmo/boltzmannbase.py
@@ -214,24 +214,26 @@ class BoltzmannBase(Theory):
         units_factors = {"1": 1,
                          "muK2": T_cmb * 1.e6,
                          "K2": T_cmb,
-                         "FIRAS": 2.7255e6}
+                         "FIRASmuK2": 2.7255e6,
+                         "FIRASK2": 2.7255
+                         }
         try:
             return units_factors[units]
         except KeyError:
             raise LoggedError(self.log, "Units '%s' not recognized. Use one of %s.",
                               units, list(units_factors))
 
-    def get_Cl(self, ell_factor=False, units="FIRAS"):
+    def get_Cl(self, ell_factor=False, units="FIRASmuK2"):
         r"""
         Returns a dictionary of lensed CMB power spectra and the lensing potential ``pp``
         power spectrum.
 
-        Set the units with the keyword ``units=number|'muK2'|'K2'|'FIRAS'``
-        (default: 'FIRAS' gives FIRAS-calibrated microKelvin^2, except for the lensing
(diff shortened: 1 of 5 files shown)


cobaya/theories/_cosmo/boltzmannbase.py
```diff
@@ -210,13 +210,28 @@ def check_no_repeated_input_extra(self):
...
         power spectrum.
 
-        Set the units with the keyword ``units='1'|'muK2'|'K2'`` (default: 'muK2',
-        except for the lensing potential power spectrum, which is always unitless).
+        Set the units with the keyword ``units=number|'muK2'|'K2'|'FIRAS'``
+        (default: 'FIRAS' gives FIRAS-calibrated microKelvin^2, except for the lensing
+        potential power spectrum, which is always unitless).
+        Note the muK2 and K2 options use the model's CMB temperature; experimental data
+        are usually calibrated to the FIRAS measurement which is a fixed temperature.
+        The default FIRAS takes CMB units = 2.7255e6 (to get result in MicroKelvin).
```
JesusTorrado, line 234: missing a "squared" at the end?

cobaya/theories/_cosmo/boltzmannbase.py
```diff
@@ -210,13 +210,28 @@ def check_no_repeated_input_extra(self):
                           "as extra arguments: %s. Please, remove one of the definitions "
                           "of each.", common)
 
-    def get_Cl(self, ell_factor=False, units="muK2"):
+    def _unit_factor(self, units, T_cmb):
```
JesusTorrado, line 213: maybe rename to _Cl_unit_factor?

JesusTorrado:
Looks good to me (just see comments above).

Also, wouldn't `FIRASmuK2` be a better name for the units?

cmbant:
Thanks @JesusTorrado , have renamed _unit_factor to be clearer. The units from that function really apply to CMB fields, e.g. the lensing-CMB cross is multiplied by _cmb_unit_factor only, auto-spectra by that squared. So in a sense the "muK2" notation is a bit unclear, so thought would make the default just "FIRAS" - but can change if you prefer.

JesusTorrado:
Looks good! Feel free to merge! (don't forget to add sth to the CHANGELOG, maybe after merge, in case of merge conflicts with my last PR merge)

JesusTorrado:
If it's not too late, I was thinking that there is some value in being explicit, so if it were just up to me I'd make it `FIRASmuK2` and would add `FIRASK2`. That lensing carries no units and cross-spectra with lensing carry a power of one is understood, I'd say. But you've got the last word there.

cmbant:
OK,  I changed it.

Issue #103, opened 2020-06-03
latex for `logprior` missing the log in polychord.py


lukashergt:
In lines 278-281 the latex entry `r"\pi_\mathrm{...}"` created for the `logprior` parameters does not include the logarithm `"\log"` or `"\ln"`:
```python
278         for p in self.model.prior:
279             f_paramnames.write("%s*\t%s\n" % (
280                 "logprior" + _separator + p,
281                 r"\pi_\mathrm{" + p.replace("_", r"\ ") + r"}"))
```

cmbant:
already fixed

PR #104, opened 2020-06-05
[PolyChord] use ppf for pc_prior instead of prior bounds

### LINKED ISSUE #101: Incorrect Kullback-Leibler divergence from PolyChord with non-uniform priors
### Example highlighting the problem:

Simple 1-dimensional example using the `one` likelihood (`one` likelihood for simplicity of the example, problem exists generally, though):
```yaml
likelihood:
  one:
params:
  a:
    prior: 
      dist: norm
      loc: 0
      scale: 1
sampler:
  polychord:
```

In the above example, we will get a log evidence of `logZ=0`, as expected for the `one` likelihood.

We would also expect a zero Kullback-Leibler divergence `D_KL=0` for the `one` likelihood, however, using the raw PolyChord data (e.g. with [anesthetic](https://github.com/williamjameshandley/anesthetic)'s `NestedSamples.ns_output()`), we incorrectly get something non-zero. 

---

### Suspected reason:

I think this happens because the prior density is not fully renormalised to [0,1] as @JesusTorrado mentions in [his answer](https://github.com/CobayaSampler/cobaya/issues/77#issuecomment-585822771) to #77:

> Indeed, instead of renormalising the prior density to [0,1] using pc_prior and then passing the pure likelihood, I am using pc_prior to just renormalise with respect to the bounds, and passing the density-part of the priors together with the likelihood. This way, we can allow for 1d priors for which we don't have an automatic renormalisation to [0,1].
 
This effectively turns the non-uniform prior into a contribution to the likelihood. This is fine for the evidence `Z`, where likelihood `L` and prior `p` enter as a product (as long as they enter as a product you can shovel prior to likelihood or likelihood to prior as you like):
```python
logZ = logsumexp(logL + logp)
```
The Kullback-Leibler divergence on the other hand has likelihood and prior entering as a product only for the first part. For the second part the likelihood enters on its own (assuming `logL` and `logp` known and `logZ` from above):
```python
D_KL = sum(        P * log(P / p)) 
     = sum(L * p / Z * log(L / Z)) 
     = sum(exp(logL + logp - logZ) * (logL - logZ))
```

---

### Potential improvement:

Could we maybe make use of the [`.ppf`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.rv_continuous.ppf.html#scipy.stats.rv_continuous.ppf) (inverse of `.cdf`) function that comes with the 1d [`scipy.stats`](https://docs.scipy.org/doc/scipy/reference/stats.html#continuous-distributions) distributions to do the conversion from unit hypercube to physical parameter? 

I'm thinking along the lines of 
1) maybe replacing line 183 in `cobaya/samplers/polychord/polychord.py`:
    ```diff
    -    self.pc_prior = lambda x: (locs + np.array(x)[self.ordering] * scales).tolist()
    +    self.pc_prior = lambda x: [prior.ppf[i](xi) for i, xi in enumerate(x)]
    ```

2) and passing the pure likelihood instead of `logposterior + self.logvolume` to `run_polychord` in lines 236-260.

I am unsure, though, how easily accessible the 1d prior distributions are in `polychord.py`. 

I should also say that I am obviously not extremely familiar with the details of Cobaya's workings, so quite possibly I have overlooked something important. Would be curious to know whether you think this might be possible, @JesusTorrado. Maybe @williamjameshandley has some suggestions, too?

lukashergt:
Hi @JesusTorrado,

I gave this a shot and it seems to be a fairly minimal change. But I'm obviously not super familiar with the internals of Cobaya, hence I am not sure whether I might be overlooking something.

This implements the changes suggested in #101 in order to get accurate Kullback-Leibler divergences also in case of non-uniform priors. 

The `pc_prior` function now uses the `.ppf` function of the `scipy.stats` 1-dimensional distributions.

Fixes #101 and #103.

### DIFF
--- a/cobaya/samplers/polychord/polychord.py
+++ b/cobaya/samplers/polychord/polychord.py
@@ -278,7 +278,7 @@ class polychord(Sampler):
             for p in self.model.prior:
                 f_paramnames.write("%s*\t%s\n" % (
                     "logprior" + _separator + p,
-                    r"\pi_\mathrm{" + p.replace("_", r"\ ") + r"}"))
+                    r"\log\pi_\mathrm{" + p.replace("_", r"\ ") + r"}"))
             for p in self.model.likelihood:
                 f_paramnames.write("%s*\t%s\n" % (
                     "loglike" + _separator + p,
--- a/cobaya/samplers/polychord/polychord.py
+++ b/cobaya/samplers/polychord/polychord.py
@@ -180,7 +180,7 @@ class polychord(Sampler):
         locs = bounds[:, 0]
         scales = bounds[:, 1] - bounds[:, 0]
         # This function re-scales the parameters AND puts them in the right order
-        self.pc_prior = lambda x: (locs + np.array(x)[self.ordering] * scales).tolist()
+        self.pc_prior = lambda x: [self.model.prior.pdf[i].ppf(xi) for i, xi in enumerate(np.array(x)[self.ordering])]
         # We will need the volume of the prior domain, since PolyChord divides by it
         self.logvolume = np.log(np.prod(scales))
         # Prepare callback function
@@ -250,7 +250,7 @@ class polychord(Sampler):
                 loglikes = np.full(self.n_likes, np.nan)
             derived = list(derived) + list(logpriors) + list(loglikes)
             return (
-                max(logposterior + self.logvolume, self.pc_settings.logzero),
+                max(loglikes.sum(), self.pc_settings.logzero),
                 derived)
 
         sync_processes()


cobaya/samplers/polychord/polychord.py
```diff
@@ -180,7 +180,7 @@ def initialize(self):
         locs = bounds[:, 0]
         scales = bounds[:, 1] - bounds[:, 0]
         # This function re-scales the parameters AND puts them in the right order
-        self.pc_prior = lambda x: (locs + np.array(x)[self.ordering] * scales).tolist()
+        self.pc_prior = lambda x: [self.model.prior.pdf[i].ppf(xi) for i, xi in enumerate(np.array(x)[self.ordering])]
```
lukashergt, line 183: The `pdf.ppf` thing might read a bit confusing. The thing is that `prior.pdf` is really a `scipy.stats.rv_continuous` class that has `.pdf`, `.logpdf` and `.ppf` as functions. Might be worth considering to rename `prior.pdf` to something else? (e.g. `prior.rv_continuous` or `prior.distributions`?)

cobaya/samplers/polychord/polychord.py
```diff
@@ -180,7 +180,7 @@ def initialize(self):
         locs = bounds[:, 0]
         scales = bounds[:, 1] - bounds[:, 0]
```
lukashergt, line 181: There might remain some possible cleaning up to do, e.g. I expect that `self.logvolume` won't be needed anymore, right? Consequently also `scales` and `locs` can be removed. Maybe we don't even need `bounds` here anymore? Meaning we could even get rid of all lines from 170 to 181. 

Or is there a different reason that PolyChord priors need to be bounded that isn't obvious to me? @williamjameshandley, maybe you can confirm whether by using the `.ppf` we indeed don't need to specify prior bounds anymore?

williamjameshandley: Certainly most other interfaces to PolyChord don't put in these bounds, so if it is necessary it might be cobaya specific. These changes all look good to me, so it's up to @JesusTorrado to approve/comment on this.

cobaya/samplers/polychord/polychord.py
```diff
@@ -278,7 +278,7 @@ def dump_paramnames(self, prefix):
             for p in self.model.prior:
                 f_paramnames.write("%s*\t%s\n" % (
                     "logprior" + _separator + p,
-                    r"\pi_\mathrm{" + p.replace("_", r"\ ") + r"}"))
+                    r"\log\pi_\mathrm{" + p.replace("_", r"\ ") + r"}"))
```
lukashergt, line 281: Would be good to add this change to #233.

PR #117, opened 2020-09-22
Add 'none' option for min_version

timothydmorton:
This allows a workaround if, e.g., the desired version of classy is old and doesn't have version access: the $CLASS_REPO_VERSION can be set to the string 'none' and the version check passes.  Might also be generally applicable elsewhere with similar env var patterns.

### DIFF
--- a/cobaya/tools.py
+++ b/cobaya/tools.py
@@ -157,7 +157,7 @@ def load_module(name, package=None, path=None, min_version=None,
         component = import_module(name, package=package)
     if path and check_path:
         check_component_path(component, path)
-    if min_version:
+    if min_version and min_version.lower() != 'none':
         check_component_version(component, min_version)
     return component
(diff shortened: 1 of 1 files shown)


JesusTorrado:
Hi @timothydmorton 

Thanks for the PR! Could you please add a warning for the case in which the check is disabled, e.g.

    if min_version and min_version.lower() != 'none':
        check_component_version(component, min_version)
    elif min_version.lower() == 'none':
        log.warning("You have disabled the minimum version check for %r.", name)

JesusTorrado:
Hi @timothydmorton 

Do you agree with my change suggestion? (Does it work for you?) Should I merge this one?

JesusTorrado:
Closing. Please let us know if this would still be useful (there is now another natural way to implement this).

PR #120, opened 2020-10-26
Camspec2020

earosenberg:
Create a new built-in likelihood class for the 2020 CamSpec likelihood, based on the 2018 version.
Includes separate likelihoods for using the usual foreground parameters (as in CamSpec 2018), parameters of power-law fits to the residuals, or both as nuisance parameters. The power-law fits should be used by default for the baseline CamSpec12.5cl cleaned likelihood.


cmbant:
Is this nearly ready for merge, or are you waiting for another paper to appear?

You can inherit bibtex files rather than repeating them, e.g. make the .bib file for the base class for all likelihoods that share a reference - this might help reduce the number of files/duplication. (we should also do this with more of the other likelihoods)

JesusTorrado:
Hi @earosenberg 

Thanks for the PR!

Is there any reason not to simplify the names, as in e.g. `planck_2020_highl_CamSpec/EE` instead of planck_2020_highl_CamSpec/EE_CS2020`?

earosenberg:
Hi both,
Thanks for the suggestions. I'll simplify the names and bib files along with a little bit more cleanup + testing, then should be ready for merge.

cmbant:
Btw, we also just added a "bibtex_file" attribute to allow easily sharing bib files between classes where they are not directly inherited- see the bibtex branch for examples of removing most duplicate bib files from the existing planck data (soon to be merged).

JesusTorrado:
Hi @earosenberg,

Would you mind adding short one-sentence descriptions to the likelihood classes as a docstring, mentioning the reference with `\cite{}`, as in https://github.com/CobayaSampler/cobaya/blob/master/cobaya/likelihoods/planck_2018_highl_CamSpec/TT.py

This is used by the `cobaya-bib` script to generate a short description of the pipeline.

earosenberg:
I wanted to keep this pull request limited to the CamSpec2020 likelihoods, but should the line `type="CMB"` be added to the_planck_2018_CamSpec_python class? This would then be inherited by CS2020.

Also, please let me know how to best distribute the data files. At the moment I have an install_options line in planck_2020_CamSpec_python.py with a dummy download URL, but this could be changed

cmbant:
You could put the data files in a GitHub download somewhere, we could probably make you a repo on the cobayasampler if needed.



cmbant:
I've added type="CMB" to the ancestor in the master branch now (thanks)

cmbant:
I made a repo at https://github.com/CobayaSampler/planck_native_data/releases/tag/v1 where data bundles can be stored in the release file, so could use that.

earosenberg:
Hi, I believe this should be ready for merge now. A comment about the name: I realized that the name I was using, `planck_2020_highl_CamSpec`, could be confused as referring to the 2020 Planck release ie NPIPE. Therefore I have renamed the likelihood to `planck_2018_highl_CamSpec2021`. Do let me know however if you think another name would be clearer or more consistent with the rest of the Cobaya likelihoods and I can change it again before merging.

cmbant:
Looks good to me. If you want to send me the zip I can add it to the repo so we can test auto-installation etc.

cmbant:
Might also be nice to add a test, e.g. with the other camspec one currently in 
https://github.com/CobayaSampler/cobaya/blob/master/tests/test_cosmo_planck_2018.py
(travis doesn't run them by default, but can be run externally or turned on)

earosenberg:
Good suggestion, thanks. Added tests along with the other CamSpec ones. I found that there is a difference in chi2 between what the test script finds and what I get from the evaluate sampler using the same point of between 0.5 and 1. True for the old 2018_CamSpec_native ones as well, not just the new ones. Couldn't track down the source of the difference though, so for now I set the test script to check for the value it was already finding

PR #122, opened 2020-10-30
Bao

Pablo-Lemos:
Just adds the Hubble distance as a BAO measurement, as in eq. (14) of https://arxiv.org/pdf/2007.08991.pdf

### DIFF
--- a/cobaya/likelihoods/_base_classes/_bao_prototype.py
+++ b/cobaya/likelihoods/_base_classes/_bao_prototype.py
@@ -244,6 +244,9 @@ class _bao_prototype(_InstallableLikelihood):
             "DA_over_rs": {
                 "angular_diameter_distance": {"z": zs.get("DA_over_rs", None)},
                 "rdrag": None},
+            "DH_over_rs": {
+                "Hubble": {"z": zs.get("DH_over_rs", None)},
+                "rdrag": None},
             "Hz_rs": {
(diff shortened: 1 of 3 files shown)


cmbant:
Weird, not sure why the PR file diff (https://github.com/CobayaSampler/cobaya/pull/122/files) is showing generic.py also changed (perhaps because a repeated pull branch?) Anyway looks fine.

What goes wrong with invcov?

Pablo-Lemos:
The strange thing is that it seems that in the repo it is currently named generic instead of bao_generic. 

The error with invcov (I do not understand why it happens): 

"likelihood 'bao.generic' does not recognize some options: 'invcov_file' (did you mean 'cov_file'|'measurements_file'?). Check the documentation for 'likelihood'."

cmbant:
Try adding invcov_file to generic.yaml?

Pablo-Lemos:
Same problem I am afraid

JesusTorrado:
Hi @Pablo-Lemos 

I have just checked explicitly that adding `invcov_file` to `generic.yaml` makes it work in your branch. Maybe you forgot to re-install after checking-out (of you are not using `pip -e`)?

Pablo-Lemos:
You're right @JesusTorrado, pushed now!

JesusTorrado:
Great! I'll merge and add a unit test for `bao.generic`

PR #127, opened 2020-11-14
Add 'make' to the apt installation list

timothydmorton:
Trying this for the first time, I found that 'make' is not installed unless requested, which means the cobaya-install of camb & classy wouldn't work without.

### DIFF
--- a/docs/cluster_amazon.rst
+++ b/docs/cluster_amazon.rst
@@ -17,7 +17,7 @@ Now install the requisites with
 
 .. code:: bash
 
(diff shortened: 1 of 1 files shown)


timothydmorton:
I also found that I needed to add `--bind-to core:overload-allowed` to the `mpirun` call in order for it to execute.

JesusTorrado:
Thanks! The `--bind-to core:overload-allowed` is necessary just since the last `openmpi` version, and the AWS script was written a little before that.

PR #128, opened 2020-11-19
Fix problem with sigma8

Pablo-Lemos:
Hello, 

I am trying to run cobaya for BAO, and I noticed that I get a sigma_8 smaller than I would expect given the rest of cosmological parameters. I checked with CosmoMC with the exact same settings, and indeed my sigma_8 in cobaya is low. I believe when using BAO, cobaya was giving me the sigma_8 at the smallest redshift at which I was evaluating BAO. This fix works, but I am sure there is a better and more efficient way of fixing the issue. 

Best, 
Pablo

### DIFF
--- a/cobaya/theories/camb/camb.py
+++ b/cobaya/theories/camb/camb.py
@@ -531,6 +531,8 @@ class camb(BoltzmannBase):
                 return derived
         # Specific calls, if general ones fail:
         if p == "sigma8":
+            intermediates.camb_params.Transfer.PK_redshifts = [0]
+            res = self.camb.get_results(intermediates.camb_params)
             return intermediates.results.get_sigma8()[-1]
         try:
             return getattr(intermediates.camb_params, p)


cmbant:
Yes, I think better to make sure z=0 is in the redshifts array before anything is originally computed. Raise an issue if you want us to look at it.

Pablo-Lemos:
Is there an easy way to make sure that z=0 is in the redshifts array? If so, I think it would be good to make that the default option. Otherwise, one could naively install the code, run it for BAO, and think they got a low sigma_8

cmbant:
You'd have to  look at how boltzmannbase.py and camb.py set up the redshift arrays from the requested inputs - may be useful going forward to understand how all this works.

cmbant:
Let's try to consolidate in https://github.com/CobayaSampler/cobaya/pull/133. Let me know if it works for you.

PR #138, opened 2021-01-13
Bibtex

cmbant:
This adds bibtex_file class option, and uses inheritance to reduce duplicated bibtex files.

Also explicitly setting halofit version for tests with latest CAMB HM2020.


JesusTorrado:
Regarding the bib part, I see one clear problem with this before merging: the text snippet before the bibtex code is different for different likelihoods (e.g. TT vs TTTEEE, DES shear vs galaxy-galaxy-lensing...) and this gets lost with these changes (unless I am missing something).

One solution is to put the snippet into a different class attribute / yaml field (e.g. "desc", as used here: https://github.com/CobayaSampler/cobaya/blob/a19d1b2a2972cdc907ce855d8685e5f8c495aecf/cobaya/cosmo_input/input_database.py#L18 ; could be promoted to conventions.py). This "desc" field could have a more general use as an optional description of the class.

JesusTorrado:
Regarding the CAMB update:
- Should there be a new "min_version" in camb.py?
- In the cosmo generator we are currently defaulting to `mead`. Should that be now `mead2020`?

cmbant:
I don't think there are any required changes in camb.
For the cosmo generator, I guess that depends on whether we want to reproduce the Planck analysis settings, or use the latest. I don't have a strong preference.

cmbant:
Regarding bibtex comment: I didn't think this was used for anything, but agree that if the idea is to have some human-readable descriptive text then it would be best to structurally separate it as different attribute (basically then the same as a standard python class comment...)

JesusTorrado:
> I don't think there are any required changes in camb.

Not even updating the min_version to 1.2? (current is 1.1.3)

> For the cosmo generator, I guess that depends on whether we want to reproduce the Planck analysis settings, or use the latest. I don't have a strong preference.

I guess when they differ enough he can have both `latest` and `Planck baseline` presets, but maybe for now let's keep 2018's `mead`, since it's the latest supported by CLASS too.

cmbant:
I think 1.1.3 is the last change that’s actually required for it to work, but can certainly update to 1.2 if you like.

JesusTorrado:
> Regarding bibtex comment: I didn't think this was used for anything, but agree that if the idea is to have some human-readable descriptive text then it would be best to structurally separate it as different attribute (basically then the same as a standard python class comment...)

Ok, I'll restore these texts and merge as soon as I am able.

JesusTorrado:
> I think 1.1.3 is the last change that’s actually required for it to work, but can certainly update to 1.2 if you like.

In that case, let's keep 1.1.3.

JesusTorrado:
I've implemented a `HasDefault.get_desc` class method that fetches the docstring of the class but can be overridden for option-dependent descriptions (see the one of mcmc).

Now we could generate the full bibliography as a list of descriptions first, and then the bibtex code itself. What do you think?

PR #140, opened 2021-01-31
Cmb forecast,  unlensed CMB, and some fixes

JesusTorrado:
- Put A. Lewis' cmb forecast dataset generator in cmblikes
- Removed underscores from base cosmo classes
- Updated cosmo model example in docs
- `model.add_requirements` does not overwrite previous calls
- unlensed Cl's now retrievable with CAMB and CLASS
- Base classes: removed _prototype where appropriate
- [...]

Still WIP.


cmbant:
For renaming things, perhaps we should do it properly by getting rid of "_prototype" (nothing, or "_base"?), and perhaps making all the classes standard CamelCase? Could also change _base_classes to base_classes (given not actually private).

Note that the renames will probably break various existing external likelihoods unless we add temporary aliases.

JesusTorrado:
Removed `_prototype`.

Re `_base_classes`, I agree it's not fully private, but the alternative to the underscore is hard-coding somewhere that it needs to be ignored. This can be some empty file such as `hidden.sth` in the corresponding folder, something in it's `__init__.py` (but makes the folder crawling a little more complicated), or hard-code it in `tools::import_all_classes`. Which one of these would you prefer? (or any other)

Regarding CamelCase, I don't think that'd be very useful of consistent: the naming of our likelihoods is effectively `folder.class`, so since folders are normally snake_case'd, we'd have things like `snake_case.CamelCase`. Not sure what's better, but I'd tend to think that it's more convenient to use either snake or camel for both (but open to discussion, of course).

Regarding external likelihoods, as a minimalist approach wrapping only the ones needed, which one do you think it's used in practice (besides `CMBlikes`)?

cmbant:
I don't there's any issue with the base classes being camel case? - e.g. we already have DataSetLikelihood, CMBlikes, etc. in some cases, so at the moment it's a mix. 

For the other build-in likelihoods I agree it would be more tricky. 
x.module.ClassName is already our general standard syntax, but to also allow just the short from x.module (where currently class names match module names), we'd have to do some sort of default module->class name cameling, or rely on an explicit "get_cobaya_class" function implementation to get the default class (as already [here](https://github.com/CobayaSampler/cobaya/blob/483ed3e07136e517451dbac10c913ce6c28b08a5/cobaya/tools.py#L194) for use by external packages, allowing Cobaya classes not be defined not in the module root and only be dynamically loaded).
I have to say I don't much like having class names matching module names as we have now, seems to break conventions and conflate qualitatively different things [and we have horrible things like theories.camb.camb.camb.camb!].

No strong preference on import_all_classes.. One could consider making abstract classes actually derived from collections.abc, and basing parse-identification on that; but I haven't yet fully understood the use cases for abc. Making the one special case "base_classes" I guess also wouldn't be too bad, or something in __init__ would be more general.

SO likelihoods are using at least "_InstallableLikelihood".

JesusTorrado:
Base classes are CamelCase now, `_base_classes` is now `base_classes`, and some (Installable, DataSet and CMBlikes) can be imported as before and show a deprecation warning.

I am open to keep discussing what to do about the rest of the classes for the next version. In principle, I'd avoid internal *magic* that makes input info name things differently from the corresponding pieces of code (modules and classes) that they reference.

Let's finish the root issue of this branch and we can discus that later.

cmbant:
in camb._get_Cl there's a spurious "raise" (I think new in this pull).

In LikelihoodInterface we now rather inconsistently have a declared _current_state, but get_current_logp is using undefined property current_state.

We also still have get_kind called with wrong argument in get_default_info().

JesusTorrado:
> in camb._get_Cl there's a spurious "raise" (I think new in this pull).

Probably leftover from testing. Solved.

> In LikelihoodInterface we now rather inconsistently have a declared _current_state, but get_current_logp is using undefined property current_state.

I see where the problem is. I'd say it makes sense to define the current state as a property. At the very least, it makes the code cleaner: even if the current state is private from the user point of view (they shouldn't modify it), we want the user to have access to it. Why not just drop LikelihoodInterface and create a `current_logp` property too for Likelihood? I don't see what LikelihoodInterface adds: there are no Likelihoods which are not LikelihoodInterface's. In fact, I would also define Theory.get_current_derived as a Theory.current_derived property.

> We also still have get_kind called with wrong argument in get_default_info().

Thanks, solved too (just removed the line, which is not needed). Tested passing a string likelihood name to get_defaut_info and it works.

cmbant:
I think LikelihoodInterface was to enable deriving likelihoods from Theory classes, e.g. see the BBN test example in cosmo_multi_theory.

cmbant:
Yes. declaring current_state rather than _current_state in LikelihoodInterface is fine.

Could certainly add current_logp and current_derived properties (the latter in Theory presumably)

JesusTorrado:
Done both. Let me know if your linter complains.

JesusTorrado:
I just got an error trying to initialise `CMBlikes` with a non-absolute `dataset_file`, and found this: https://github.com/CobayaSampler/cobaya/blob/f8cc98a432021b151246cb6f644aebd7895f921a/cobaya/likelihoods/base_classes/DataSetLikelihood.py#L45

Reading the comment after the else below, wouldn't it make sense to keep the relative path as long as there are not install options? so `if os.path.isabs(self.dataset_file) or not self.get_install_options()` or something like that. (The dataset_file path could be made internally absolute after the if anyway

cmbant:
I think the assumption was if is relative then it's relative to the class location - e.g. if people want to package a derived class and .dataset files together in a folder/package somewhere?

PR #144, opened 2021-02-24
Sigma 8 z

Pablo-Lemos:
I needed to use sigma_8(z) from cobaya, and I thought it could be generally useful. The current version only allows to use sigma_R, with R in units h^{-1}, so there is no straightforward way of getting sigma_8.


cmbant:
Thanks! Currently I think not implemented for Class. I added 
https://github.com/CobayaSampler/cobaya/pull/145
which would allow @abstract decorator to raise error or be ignored if used with class (or any other non-implementing descendant)- or could implement.

JesusTorrado:
Looks good!

@cmbant what do you mean by "ignored" there?

cmbant:
The dependency analyser will not assign to get_x methods where get_x is @abstract.

JesusTorrado:
Then I guess the provider will complain that it cannot find anyone that prodives sigma8 in this case. I guess this is a better behaviour than complaining about CLASS precisely, since there could be another theory code.

JesusTorrado:
Ok, it's clear now. Forget my last comment.

CLASS implementation for this and other stuff will hopefully come in the next few weeks, depending on how busy Julien et al are.

cmbant:
@Pablo-Lemos please could you rebase from master and add @abstract decorator on get_sigma_8_z in the Boltzmann base, then I think good to go if all OK.

Pablo-Lemos:
Done! (I think)

Issue #90, opened 2020-05-11
Example in the paper does not work when formulated in terms of params


JesusTorrado:
[citing AL]

Note this plausible looking rewrite in terms of parameters does not work (nor generate a helpful error message):

    class ToPolarCoords(Theory):
        params = {'x': None, 'y': None}

        def get_can_provide_params(self):
            return ['r', 'theta']

        def calculate(self, state, want_derived=True, **params_values):
            x = params_values['x']
            y = params_values['y']
            state['derived'] = {'r': np.sqrt(x ** 2 + y ** 2), 'theta': state['derived'] }

    class GaussianRingLikelihood(Likelihood):
        params = {'mean_radius': None, 'std_radius': None, 'r': None}

        def logp(self, **params_values):
            r = params_values['r']
            mean_radius = params_values['mean_radius']
            std_radius = params_values['std_radius']
            return stats.norm.logpdf(r, loc=mean_radius, scale=std_radius)

Not sure what Cobaya is getting as the PDF for 'r': None in this case, where it has no information?

PR #147, opened 2021-03-12
component params converted to requirements if not explicitly sampled

cmbant:
Possible fix to make example in https://github.com/CobayaSampler/cobaya/issues/90 work


JesusTorrado:
Thanks! Is this ready to review and merge?

cmbant:
Yes I thinks so

JesusTorrado:
Neat solution! Added a comment for us to able to understand it in the future.

PR #150, opened 2021-03-17
Reducing the overhead of MCMC

JesusTorrado:
- [x] add numpy cache to collection --> now ~3-4x faster than get_proposal (more without numba)
- [x] check_convergence: compute bounds only if Rminus1 bounds test will be performed (affects dim ~< 10) --> 30% faster
- [ ] ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## 3.0.5 – 2021-YY-ZZ
 
+### MCMC
+
+- Overhead reduced by at least 40%, thanks to caching in Collection
+
 ### Minimize
 
 - `PyBOBYQA` updated to 1.2, and quieter by default.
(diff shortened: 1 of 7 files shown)


JesusTorrado:
Done on my side. Increasing the cache size beyond the current level does not have a noticeable effect.

@cmbant Any other speed improvement you had in mind that makes sense to try at this time?

cmbant:
Sounds good, I don't think so - I thought we were already at the level where even a fairly trivial Gaussian likelihood would always dominate the time? 

I did wonder if _cache_add_row couldn't just store the arrays (like OneSamplePoint), with the loops to resolve names done just once at the time of cache dump [I guess for timing it doesn't make too much difference, as the point is just to avoid slow Pandas].

JesusTorrado:
The posterior dominates, but not by much: ~1.5x larger than the overhead before, and ~2.5x now, according to the overhead test in `test_mcmc.py`, so it's a substantial improvement at little cost.

I also thought of doing the dict lookups and checks when cache-dumping, but I think it would make it harder to debug if it come to it (e.g. users doing their own implementation and for any reason at all changing the order in which they add values to the dicts passed to Collection.add).

I'll remove some redundant decorators and merge then.

PR #152, opened 2021-03-19
More general input derived parameter dependencies

cmbant:
Currently things like
```

  xx: 
    value: 12
    drop: True
  xy:
    value: "lambda xx: xx*2"
    drop: True
  x:
    value: "lambda xy, xx: xx*xy"  
```
generate an error that "xy" is not used by any likelihood (assuming it isn't). Ideally we'd support nested dependencies, but I think currently not supported, so  this pull just changes the way the existing nested dependency error is raised. Not sure if that error is supposed to be catching other cases not captured by this.

### DIFF
--- a/cobaya/input.py
+++ b/cobaya/input.py
@@ -187,10 +187,8 @@ def update_info(info):
             # Reserved attributes not necessarily already in default info:
             reserved = {_external, _class_name, _provides, _requires, partag.renames,
                         _input_params, _output_params, _component_path, _aliases}
-            options_not_recognized = (set(input_block[component])
-                                      .difference(reserved)
-                                      .difference(set(updated[component]))
-                                      .difference(set(annotations)))
+            options_not_recognized = set(input_block[component]).difference(
+                chain(reserved, updated[component], annotations))
             if options_not_recognized:
                 alternatives = {}
                 available = (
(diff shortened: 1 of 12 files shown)


cmbant:
Now trying to make nested lambda dependencies like the example above just work

cmbant:
@JesusTorrado, do I recall correctly that drop is now always redundant as long as not using any agnostic theory - i.e. I guess only needed when used with Class, or does it have other uses?

cmbant:
Notes: parameterization._derived_dependencies is currently not used

@Pablo-Lemos I think this should fix the issue you had trying to use omegab as an input parameter, if you have a chance to test it let me know if it still doesn't work

JesusTorrado:
> @JesusTorrado, do I recall correctly that drop is now always redundant as long as not using any agnostic theory - i.e. I guess only needed when used with Class, or does it have other uses?

You may want to give a name to a parameter that coincides with one understood by e.g. CAMB (maybe accidentally: CAMB understands so many parameters that you may not know that your defined one shares a name with a CAMB one, and you'd get a strange error in that case if you don't use `drop`).

cmbant:
OK thanks. I've re-jigged this a bit, so dropped parameters are now included in input_params, and only explicitly removed when assigning to theory/likelihoods (so priors can now depend on dropped and undropped input and derived parameters, and simplifying a few things).

PR #160, opened 2021-04-07
Importance sampling: updates, no unnecessary theory computation, mpi, using CosmoMC input chains

cmbant:
Attempt to allow simple importance sampling over cosmomc chains, after running a conversion script to get an input .yaml. There are also a few general tidies/optimizations/refactors which should probably be included anyway (and some others for changes in function_deps branch). The yaml script is included in example below. It appears to work, at least in the case of adding a likelihood.

Sample code to importance sampling Planck lensing chain with BAO:

```
def cosmomc_root_to_cobaya_info_dict(root, derived_to_input=()):
    names = ParamNames(root + '.paramnames')
    if os.path.exists(root + '.ranges'):
        ranges = ParamBounds(root + '.ranges')
    else:
        ranges = None
    d = {}
    info = {'params': d}
    for name in names.names:
        if name.name.startswith('chi2_') and not name.name.startswith('chi2__'):
            if name.name == 'chi2_prior':
                continue
            name.name = name.name.replace('chi2_', 'chi2__')
        d[name.name] = {'latex': name.label}
        if name.renames:
            d[name.name]['renames'] = name.renames
        if name.isDerived:
            if name.name not in derived_to_input:
                d[name.name]['derived'] = True
            else:
                name.isDerived = False
        if ranges and name.name in ranges.names:
            if name.isDerived:
                d[name.name]['min'] = ranges.getLower(name.name)
                d[name.name]['max'] = ranges.getUpper(name.name)
            else:
                d[name.name]["prior"] = [ranges.getLower(name.name),
                                         ranges.getUpper(name.name)]
    if ranges:
        d.update(ranges.fixedValueDict())
    if names.numberOfName('As') == -1 and names.numberOfName('logA') != -1:
        d['As'] = {'latex': r'A_\mathrm{s}', 'value': 'lambda logA: 1e-10*np.exp(logA)'}
    if names.numberOfName('cosmomc_theta') == -1 and names.numberOfName('theta') != -1:
        d['cosmomc_theta'] = {'latex': r'\theta_{\rm MC}',
                              'value': 'lambda theta: theta/100'}

    # special case for CosmoMC (e.g. Planck) chains
    if os.path.exists(root + '.inputparams'):
        inputs = IniFile(root + '.inputparams')
        for key, value in inputs.params.items():
            if key.startswith('prior['):
                if 'prior' not in info:
                    info['prior'] = {}
                param = key[6:-1]
                if param in d:
                    mean, std = (float(v.strip()) for v in value.split())
                    if not names.parWithName(param).isDerived:
                        info['prior'][param + '_prior'] = \
                            "lambda %s: stats.norm.logpdf(%s, loc=%g, scale=%g)" % (
                                param, param, mean, std)
            if key.startswith('linear_combination['):
                param = key.replace('linear_combination[', '')[:-1]
                prior = inputs.params.get('prior[%s]' % param, None)
                if prior:
                    weights = inputs.params.get('linear_combination_weights[%s]' % param,
                                                None)
                    if not weights:
                        raise ValueError(
                            'linear_combination[%s] prior found but not weights' % param)
                    weights = [float(w.strip()) for w in weights.split()]
                    inputs = value.split()
                    if 'prior' not in info:
                        info['prior'] = {}
                    mean, std = (float(v.strip()) for v in prior.split())
                    linear = "".join(
                        "%+g*%s" % (_w, _p) for _w, _p in zip(weights, inputs))
                    info['prior']['SZ'] = \
                        "lambda %s: stats.norm.logpdf(%s, loc=%g, scale=%g)" % (
                            ",".join(inputs), linear, mean, std)
    if os.path.exists(root + '.likelihoods'):
        info_like = {}
        info['likelihood'] = info_like
        with open(root + '.likelihoods', 'r') as f:
            for line in f.readlines():
                if line.strip():
                    like = line.split()[2]
                    info_like[like] = None
    else:
        print('You need to mention in the yaml a likelihood block with with "name: None"'
              'for each likelihood in the input chain')

    info['output'] = root
    return info


def test_likelihood(_self, omegabh2):
    return 1.2


info = cosmomc_root_to_cobaya_info_dict(
    r'C:\Tmp\CobayaChains\cosmomc\base_lensing_lenspriors')
post = {}
info['post'] = {'add': post, 'suffix': 'post', 'skip': 0.9,
                'packages_path': r'C:\Work\Dist\git\cobaya\tests\modules'}
#post['likelihood'] = {
#    "test_likelihood": {'external': test_likelihood}}
post['likelihood'] = {'bao.sdss_dr12_consensus_bao': None}
post['theory'] = {'camb': None}
info['stop_at_error'] = True
info['debug'] = False
info['force'] = True

from cobaya.post import post

print(info)
post(info)
`
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -107,7 +107,7 @@ script:
   - export COBAYA_PACKAGES_PATH="../packages"
   # mpi tests [not sure why it hangs in py3.6]
   - if [[ "$GCC_VERSION" != "6" ]]; then
-      mpiexec -np 2 --mca orte_base_help_aggregate 0 --mca btl ^openib --oversubscribe pytest -x tests/test_mcmc.py tests/test_post.py --no-flaky-report ;
+      mpiexec -np 2 --mca orte_base_help_aggregate 0 --mca btl ^openib --oversubscribe pytest -x -m mpi tests/ --no-flaky-report ;
     fi;
   # General tests:
   - cobaya-install polychord --debug
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,8 +65,6 @@ jobs:
       dist: focal
       addons:
         apt:
-          sources:
-            - ubuntu-toolchain-r-test
           packages:
             - gcc-10
             - gfortran-10
@@ -107,6 +105,10 @@ before_install:
 
 script:
   - export COBAYA_PACKAGES_PATH="../packages"
+  # mpi tests [not sure why it hangs in py3.6]
+  - if [[ "$GCC_VERSION" != "6" ]]; then
+      mpiexec -np 2 --mca orte_base_help_aggregate 0 --mca btl ^openib --oversubscribe pytest -x tests/test_mcmc.py tests/test_post.py --no-flaky-report ;
+    fi;
   # General tests:
   - cobaya-install polychord --debug
   - pytest tests/ -k "not cosmo" --skip-not-installed --no-flaky-report
--- a/TODO.md
+++ b/TODO.md
@@ -6,12 +6,12 @@
 
 # cosmetic/consistency/speed
 
+## make portalocker a requirement?
 ## version attribute should be in all components not just theory (samplers can have versions) [done for samplers; missing: likelihoods]
-## post: use MPI, and TODO's in code
+## post: TODO's in code
 ## In the docs "Bases" (and UML diagram) not hyperlinked correctly (not sure how to fix)
 ## Make numba a requirement?
 ## dump log info along with each chain file if saving to file (currently in stdout)
-## Faster Collections for MCMC: numpy cache for merging OnePoint into Collection, `out_update` method would take care of flushing into the Pandas table.
 ## PolyChord: check overhead
 ## PolyChord: lower dimension of tests?
 
@@ -42,5 +42,4 @@
 * Maybe remove distinction between input parameters and requirements, so that `calculate`/`logp` takes both of them, which would be prepared by `check_cache_and_compute`. This would simplify the code a bit (in particular the part about input parameters that can be requirements, e.g. YHe) and makes all likelihood automatically callable outside a `Model` feeding requirements by hand. Problem: to prepare requirements we need arguments (e.g. units, `ell_factor` for Cl's) which are not passed to `must_compute`.
 * AL: We should perhaps allow all requirements to also be arrays of Requirement or (key, option) tuples (this is already done internally for combining requirements, but currently no component can individually return two requests with the same key but different options because dicts have unique keys).
 * Provider: it should be possible to save retrieving methods at initialisation so that everything (params, results, methods) can be retrieved with Provider.get(**args). Maybe it is interesting?
-## Implement file locks for testing correct MPI configuration
-## Add MPI tests for convergence and resuming (see pytest-mpi)
+## Add MPI tests for convergence, resuming and post (see pytest-mpi)
--- a/cobaya/bib.py
+++ b/cobaya/bib.py
@@ -98,9 +98,6 @@ def prettyprint_bib(descs, bibs):
 
 # Command-line script
 def bib_script():
-    from cobaya.mpi import is_main_process
-    if not is_main_process():
-        return
     warn_deprecation()
     # Parse arguments and launch
     import argparse
@@ -116,11 +113,13 @@ def bib_script():
                         help=("If component name given, "
                               "kind of component whose bib is requested: " +
                               ", ".join(['%s' % kind for kind in kinds]) + ". " +
-                              "Use only when component name is not unique (it would fail)."))
+                              "Use only when component name is not unique "
+                              "(it would fail)."))
     arguments = parser.parse_args()
     # Case of files
     are_yaml = [
-        (os.path.splitext(f)[1] in _yaml_extensions) for f in arguments.components_or_files]
+        (os.path.splitext(f)[1] in _yaml_extensions) for f in
+        arguments.components_or_files]
     if all(are_yaml):
         infos = [load_input(f) for f in arguments.components_or_files]
         print(prettyprint_bib(*get_bib_info(*infos)))
--- a/cobaya/collection.py
+++ b/cobaya/collection.py
@@ -125,16 +125,21 @@ class Collection(BaseCollection):
                     self._out_load(skip=onload_skip, thin=onload_thin)
                     if load:
                         self.columns = list(self.data.columns)
-                        loaded_chi2_names = [name for name in self.columns if
-                                             name.startswith(_chi2 + _separator)]
-                        if _chi2 + _separator + 'prior' in loaded_chi2_names:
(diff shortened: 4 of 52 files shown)


cmbant:
Now also updated and changed handling of theories, so no theories are recomputed unless their outputs are actually needed for something (precomputed derived parameters are available to likelihoods from the input samples). To do this first assign parameters as normal, but with likelihood input parameters now all the available precomputed parameters; theory parameters are however only those normally assigned to theories, so that they will run OK if recomputed. Theories which are not needed for anything are then removed from computed component list, so not recomputed. This should allow e.g. quick use of sigma8 from likelihoods.

misharash:
I had an issue with minimizer, using older version of this branch. I searched for best fit in CMB run, and `chi-sq` at the top of the file was somehow smaller than `chi2__CMB` by some weird amount close to 30. The log-likelihoods from `polychord` output are more consistent with the latter. I'm currently checking whether it still happens with latest commits.

misharash:
I've run post-processing of CMB (Planck lowl TT/EE, highl plik lite and lensing) run with CLASS on this branch, trying to get some additional derived parameters out of CLASS (also passing more extra params), not recomputing likelihoods, and had several issues.
1. tools.py, line 399: string has no items(). Turned out that base='v2.9.4' (string, version of CLASS I use). Workaround: put `if type(base) == str: return base` above.
2. post.py, line 425: None has no check_error(). Probably because I've run original chain with 4 MPI processes 4 threads each and post-processing with only one. Workaround: comment out the line.
3. `getdist` confuses the order of new derived params with chi2__CMB when it reads the post-processed chain. In samples.paramNames.names, chi2__CMB comes before the new derived params, while in data array it comes after. Can be fixed by changing order in any of those.

cmbant:
Thanks, do you know the traceback for 1?

For 2 indeed it should be "if mpi.process_state": ...

For 3 can you share the chain files to reproduce?

I have fixed quite a few more bugs (and more coming shortly..), it may be best to use branch which has all the latest https://github.com/CobayaSampler/cobaya/pull/163

misharash:
I've switched to #163, but currently all issues I've described appear there too.
For 1:
```
Traceback (most recent call last):
  File "/usr/bin/cobaya-run", line 33, in <module>
    sys.exit(load_entry_point('cobaya==3.0.2', 'console_scripts', 'cobaya-run')())
  File "/home/misharash/Code/cobaya/cobaya/cobaya/run.py", line 228, in run_script
    run(info, **arguments.__dict__)
  File "/home/misharash/Code/cobaya/cobaya/cobaya/mpi.py", line 300, in wrapper
    return func(*args, **kwargs)
  File "/home/misharash/Code/cobaya/cobaya/cobaya/run.py", line 92, in run
    post(info)
  File "/home/misharash/Code/cobaya/cobaya/cobaya/mpi.py", line 300, in wrapper
    return func(*args, **kwargs)
  File "/home/misharash/Code/cobaya/cobaya/cobaya/post.py", line 239, in post
    recursive_update(theory_info, added_theory.pop(theory))
  File "/home/misharash/Code/cobaya/cobaya/cobaya/tools.py", line 393, in recursive_update
    base[update_key] = recursive_update(
  File "/home/misharash/Code/cobaya/cobaya/cobaya/tools.py", line 398, in recursive_update
    for k, v in (base or {}).items():
AttributeError: 'str' object has no attribute 'items'
```
For 3, the files are [here](https://github.com/CobayaSampler/cobaya/files/6366891/issue3.zip), including the notebook illustrating the problem. `chi2__CMB` comes before the group of `_rec` quantities in the parameter info array, but in data columns it comes after.

cmbant:
Thanks, will have a look. For (1), what is the dict being merged - there's a version dictionary in there? (or just a dump of the two dicts would be enough to reproduce - I can't run class easily)

misharash:
At the first iteration of this recursive update,
`base = {'path': None, 'speed': 0.2, 'stop_at_error': False, 'extra_args': {'non linear': 'halofit', 'recombination': 'recfast', 'N_ur': 3.046}, 'use_renames': False, 'renames': {'As': 'A_s', 'ns': 'n_s', 'nrun': 'alpha_s', 'nrunrun': 'beta_s', 'nt': 'n_t', 'ntrun': 'alpha_t', 'rdrag': 'rs_drag', 'omegak': 'Omega_k', 'omegal': 'Omega_Lambda', 'w': 'w0_fld', 'wa': 'wa_fld', 'omegabh2': 'omega_b', 'omegab': 'Omega_b', 'omegach2': 'omega_cdm', 'omegac': 'Omega_cdm', 'omegam': 'Omega_m', 'omegan': 'Omega_nu', 'tau': 'tau_reio', 'zrei': 'z_reio', 'deltazrei': 'reionization_width', 'helium_redshift': 'helium_fullreio_redshift', 'helium_delta_redshift': 'helium_fullreio_width', 'yhe': 'YHe', 'yheused': 'YHe'}, 'version': 'v2.9.4', 'input_params': ['A_s', 'n_s', '100*theta_s', 'omega_b', 'omega_cdm', 'tau_reio'], 'output_params': ['H0', 'Omega_m', 'Omega_Lambda', 'YHe', 'z_reio', 'sigma8', 'age', 'rs_drag']}`,
`update = {'path': None, 'speed': 0.2, 'stop_at_error': False, 'extra_args': {'compute damping scale': 'y'}, 'use_renames': False, 'renames': {'As': 'A_s', 'ns': 'n_s', 'nrun': 'alpha_s', 'nrunrun': 'beta_s', 'nt': 'n_t', 'ntrun': 'alpha_t', 'rdrag': 'rs_drag', 'omegak': 'Omega_k', 'omegal': 'Omega_Lambda', 'w': 'w0_fld', 'wa': 'wa_fld', 'omegabh2': 'omega_b', 'omegab': 'Omega_b', 'omegach2': 'omega_cdm', 'omegac': 'Omega_cdm', 'omegam': 'Omega_m', 'omegan': 'Omega_nu', 'tau': 'tau_reio', 'zrei': 'z_reio', 'deltazrei': 'reionization_width', 'helium_redshift': 'helium_fullreio_redshift', 'helium_delta_redshift': 'helium_fullreio_width', 'yhe': 'YHe', 'yheused': 'YHe'}, 'version': None}`
Seems like it gets to `'version'` and crashes, because it is not a dictionary but string in old info, while in new info it's None which is converted to empty dictionary, invoking the recursion. Probably you should check the base value too before calling `recursive_update`?

cmbant:
Thanks 1+2 should be fixed in https://github.com/CobayaSampler/cobaya/tree/info_dicts.
Getdist issue has more possible solutions.

cmbant:
Hopefully 3 is now also fixed in https://github.com/CobayaSampler/cobaya/tree/info_dicts (the post updated.yaml is a merged file - seems much safer than leaving it to external programs like getdist to figure out how thing are combined).

misharash:
Indeed, all three are now fixed in [info_dicts](https://github.com/CobayaSampler/cobaya/tree/info_dicts). Thank you!

PR #175, opened 2021-05-12
Consolidated Updates 2

cmbant:
Minimize starts plus
* clik update
* typing and code clean
* mcmc acceptance rate skips first chunk
* Collection -> SampleCollection
* Support CamelCase class names in similarly-named module; file_base_name attribute for bib and yaml


cobaya/likelihoods/base_classes/planck_clik.py
```diff
@@ -52,8 +52,8 @@ def initialize(self):
             raise LoggedError(
                 self.log, "No path given to the Planck likelihood. Set the "
                           "likelihood property 'path' or the common property "
-                          "'%s'.", _packages_path)
-        clik = is_installed_clik(path=self.path_clik, allow_global=allow_global)
+                          "'%s'.", "packages_path")
```
JesusTorrado, line 55: nor grabbed from conventions?

cobaya/likelihoods/base_classes/planck_clik.py
```diff
@@ -292,6 +289,7 @@ def execute(command):
     if _clik_verbose:
         process = Popen(command, stdout=PIPE, stderr=STDOUT)
         out = []
+        assert process.stdout
```
JesusTorrado, line 292: undhandled expception?

cobaya/likelihoods/base_classes/planck_clik.py
```diff
@@ -360,7 +345,7 @@ def install_clik(path, no_progress_bars=False):
 
 def get_product_id_and_clik_file(name):
     """Gets the PLA product info from the defaults file."""
-    defaults = get_default_info(name, kinds.likelihood)
+    defaults = get_default_info(name, "likelihood")
```
JesusTorrado, line 348: Not grabbed from conventions now?

tests/test_prior_external.py
```diff
@@ -8,25 +8,25 @@
...
 """
 # Local
-from cobaya.conventions import _prior
 from .common_external import info_string, info_callable, info_mixed, info_import
 from .common_external import body_of_test
 
 
 def test_prior_external_string(tmpdir):
-    body_of_test(info_string, _prior, tmpdir)
+    body_of_test(info_string, "prior", tmpdir)
```
JesusTorrado, line 20: Should be grabbed from conventions, and next ones too

JesusTorrado:
Is there any reason that I don't know of for dropping all the "callbacks" to names defined in conventions?

cobaya/likelihoods/base_classes/planck_clik.py
cmbant: I've tried to remove all non-private variables with underscore names for style consistency and to reduce the very verbose import statements. Some I renamed, maybe we should do this for packages path. 
The idea of using TypedDict was that the static consistency checking that originally came from using imported _var names is now available via the typed dict.

cobaya/likelihoods/base_classes/planck_clik.py
cmbant: Yes, it would have given an exception before too - but stdout should not be none  (the assert just shuts up static checking warnings).

tests/test_prior_external.py
cmbant: There are hundreds of these..

cobaya/likelihoods/base_classes/planck_clik.py
cmbant: This is a bit annoying.. we could certainly still use a kinds namedtuple (btw, I think StrEnum is coming in py10, which might be what we really want) - however none of the usual static checkers seem to then be happy using kinds.likelihood in a typed dict (needs the finalized literal "likelihood").

cobaya/likelihoods/base_classes/planck_clik.py
JesusTorrado: > I've tried to remove all non-private variables with underscore names for style consistency and to reduce the very verbose import statements.

I really do not think this is worth losing the ability to refactor much more easily by changing definitions in a single place, and it is in fact duplicated code in the form of duplicated string definitions. If verbosity and underscores are the problem, I would propose something like `import cobaya.conventions as conv`, and then use `conv.[whatever]`.

> The idea of using TypedDict was that the static consistency checking that originally came from using imported _var names is now available via the typed dict.

Not for every value, right? because not everything is a type (neither it should be).

cobaya/likelihoods/base_classes/planck_clik.py
JesusTorrado: Static checkers can usually be configured to ignore selected warnings, I guess. I really don't think having a static checker not to complain is worth the code duplication and harder refactoring. I understand there are hundreds of this and I am not asking that they be reverted just now, but I'd honestly like to go back to the single-place definition, maybe with the shorter import as suggested above.

JesusTorrado: (I don't mean changing them before the merge, of course, but at some point later. I am still slowly understanding the new code.)

PR #185, opened 2021-06-25
Added Boss DR16 likelihood

Pablo-Lemos:


cmbant:
Thanks! If you are using PyCharm, you must not be auto-formatting the source to get these whitespace inconsistency errors? - see
https://github.com/CobayaSampler/cobaya/blob/master/DEVEL.rst

Pablo-Lemos:
Done! I have also updated input_database

cmbant:
I guess this needs https://github.com/CobayaSampler/bao_data/pull/1 to be merged (and make a new release?) so the data are there before tests will pass, I think @JesusTorrado would have to merge as I don't seem to have permissions on that repo.

cobaya/cosmo_input/input_database.py
```diff
@@ -479,10 +479,16 @@
 
 like_bao = {none: {},
             'BAO_planck_2018': {
```
JesusTorrado, line 481: I think that for this we should create a new BAO set set to the latest comprehensive data combination available (`BAO_last`, or sth like that?) and preserve the Planck 2018 BAO set for archival purposes. What do you both think?

JesusTorrado:
Thanks, @Pablo-Lemos !

Could you update in the BAO class definition "github_release" to "v2.0" please?

cobaya/cosmo_input/input_database.py
Pablo-Lemos: Yes that sounds like a good idea

Pablo-Lemos:
Done!

cmbant:
@JesusTorrado  OK to merge?

PR #188, opened 2021-07-20
Pandas 1.3 changes

cmbant:
Pandas 1.3 seems to have royally messed up a bunch of things, with series objects apparently returning an object-type array that doesn't work with various numpy functions any more. I've attempted to change to use explicit .to_numpy(dtype=np.float64). Seems to work except bleeding edge conda-forge.

@JesusTorrado, we should probably do a new version ASAP as the pandas changes break many tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,6 +71,8 @@ jobs:
         - PYDIST="ANACONDA"
         - ANACONDA_CHANNEL="conda-forge"
       python: "3.9"
+  allow_failures:
+    - name: "Latest: gcc-11, Python 3.9 Anaconda (conda-forge)"
 
 before_install:
   # Configure right compiler versions
--- a/DEVEL.rst
+++ b/DEVEL.rst
@@ -7,7 +7,7 @@ This document gathers some notes about the development flow, release checklist, 
 
   This document may not always be up to date
 
-git`` development model
+``git`` development model
 -------------------------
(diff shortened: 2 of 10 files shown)


cmbant:
The bleeding edge travis fail seem to be due to 

```
Lbuild/temp.linux-x86_64-3.9 -o _configtest
  /home/travis/miniconda/envs/test-environment/compiler_compat/ld: warning: libopen-rte.so.40, needed by /usr/lib/x86_64-linux-gnu/openmpi/lib/libmpi.so, not found (try using -rpath or -rpath-link)
  /home/travis/miniconda/envs/test-environment/compiler_compat/ld: warning: libopen-pal.so.40, needed by /usr/lib/x86_64-linux-gnu/openmpi/lib/libmpi.so, not found (try using -rpath or -rpath-link)
  /home/travis/miniconda/envs/test-environment/compiler_compat/ld: warning: libm.so.6, needed by /usr/lib/x86_64-linux-gnu/openmpi/lib/libmpi.so, not found (try using -rpath or -rpath-link)
  /home/travis/miniconda/envs/test-environment/compiler_compat/ld: warning: libhwloc.so.15, needed by
```
etc. I guess pending any ideas (more general solution to the LD_PRELOAD currently used for similar issue with libgfortran) we go back to allow failures

Issue #110, opened 2020-06-26
Pass list parameters to CLASS 


tkarwal:
CLASS takes input for certain parameters as a list in its .ini files. Egs. include
m_ncdm = 0.04, 0.04, 0.04
scf_parameters = 10.0, 0.0, 0.0, 0.0, 100.0, 0.0
and many others. 

Without rewriting significant portions of CLASS, it would be nice to be able to pass these parameters through Cobaya to CLASS as well as have the option to vary them through an MCMC. By that I mean, be able to do something like: vary m_ncdm[0] and m_cdm[1] independently as separate MCMC parameters but fix m_ncdm[2]. 

One way of doing this, drawing form MontePython, is to define these parameters as 
m_ncdm__1
m_ncdm__2
m_ncdm__3 
in the cobaya .yaml file. I have added code to cobaya/theories/classy.py to read in all the parameters the sampler passes, and convert any with "__i"  to the correct format just before passing them on to CLASS. 
These then behave as independent parameters for the sampler, but CLASS reads them in correctly as 
m_ncdm = m_ncdm[0], m_ncdm[1], m_ncdm[2] 
where m_ncdm[i] are values and the above RHS is passed as a string. 

This is done by modifying the function 'calculate' in cobaya/theories/classy/classy.py to read (full function not printed here):

    def calculate(self, state, want_derived=True, **params_values_dict):
        # Set parameters
        ### new code to read in list parameters
        list_params = [key[:-3] for key in params_values_dict.keys() if '__1' in key.lower()]
        # this looks for any parameters defined as __1, i.e. the beginning of list params
        # and saves just the name minus the '__1' in a list
        if list_params:
            self.log.debug('Found params that have __1 in them:')
            self.log.debug(list_params)
        #else:
            #self.log.debug('Found no params that have __1 in them\n')
        # print statements to check that this works.
        for orig_name in list_params:
            list_param_values = [value for key,value in params_values_dict.items() if re.search(r'%s__([1-9])' % orig_name, key)]
            # Find the values of the others in the list parameter indices and stick them all together into one list of numbers
            values_str = ', '.join(['%g' % value for value in list_param_values])
            params_values_dict[orig_name] = values_str
            self.log.debug('Resetting list parameter %s = %s', orig_name, values_str)
            # Define a new input param with new name without the '__1' at the end as a string of comma-separated numbers
            for index in range(1, len(list_param_values)+1):
                del params_values_dict[orig_name + '__%i' % index]
[...]

PR #191, opened 2021-08-12
Simple implementation for vector parameters

JesusTorrado:


JesusTorrado:
@cmbant do you want to take a look before merge? Should be done.

It's just documentation and better error control.

Only actual change needed is in getdist (see https://github.com/cmbant/getdist/pull/68)

Minimal working example:

    from scipy import stats


    def gauss_band(x_vector):
        return stats.norm.logpdf(
            x_vector[0] + x_vector[1], loc=1, scale=0.02)

    # NB: don't forget the 'drop: True' in the individual parameters
    #     and 'derived: False' in the vector parameters

    info = {
        "likelihood": {"band": gauss_band},
        "params": {
            "x": {"prior": {"min": 0, "max": 2}, "proposal": 0.1, "drop": True},
            "y": {"prior": {"min": 0, "max": 2}, "proposal": 0.1, "drop": True},
            "x_vector": {"value": lambda x, y: [x, y], "derived": False}},
        "sampler": {"mcmc": None}}

    info["force"] = True
    info["output"] = "chains/band"

    from cobaya.run import run
    updated_info, sampler = run(info)

cmbant:
Interesting that just works, looks OK to me though have not tested.

JesusTorrado:
Yeah, surprisingly easy. I vaguely remember seeing this applied to some weird "vector is string form" of CLASS by @nataliehogg in some issue that I cannot find at the moment (so @nataliehogg, if you see this and can confirm, I am happy to give you credit in the changelog).

As for mergning, I would wait until the next GetDist release, since it's a requirement.

lukashergt:
> Yeah, surprisingly easy.

To a big part that is down to the nice flexible input setup of Cobaya. Glad [#110](https://github.com/CobayaSampler/cobaya/issues/110#issuecomment-652333489) proved helpful.

PR #193, opened 2021-08-14
LogPosterior as dataclass and a couple of Collection improvements

JesusTorrado:
- [x] LogPosterior is now a dataclass: allows for automatically setting summed logpriors/likelihoods at init (needs `dataclasses` package if Python 3.6 used; package has no effect for 3.7+)
- [ ] ...

### DIFF
--- a/cobaya/collection.py
+++ b/cobaya/collection.py
@@ -190,10 +190,8 @@ class SampleCollection(BaseCollection):
                         logpriors=logpriors, loglikes=loglikes)
 
     def add_log_posterior(self, values: np.ndarray, results: LogPosterior, weight=1):
-        logprior_sum = sum(results.logpriors)
-        loglike_sum = sum(results.loglikes)
-        assert np.isclose(logprior_sum + loglike_sum, results.logpost)
-        self._cache_add(values, (results.logpost, logprior_sum, loglike_sum),
+        assert np.isclose(results.logprior + results.loglike, results.logpost)
+        self._cache_add(values, (results.logpost, results.logprior, results.loglike),
                         derived=results.derived, weight=weight,
                         logpriors=results.logpriors, loglikes=results.loglikes)
 
--- a/cobaya/model.py
+++ b/cobaya/model.py
@@ -6,13 +6,14 @@
 
 """
 # Global
+import os
+import dataclasses
+import numpy as np
 from contextlib import contextmanager
 from copy import deepcopy
 from itertools import chain
 from typing import NamedTuple, Sequence, Mapping, Iterable, Optional, \
     Union, List, Any, Dict, Set
-import numpy as np
-import os
 
 # Local
 from cobaya.conventions import overhead_time, debug_default, get_chi2_name, \
@@ -44,12 +45,24 @@ def timing_on(model: 'Model'):
             model.set_timing_on(False)
 
 
-# Log-posterior namedtuple
-class LogPosterior(NamedTuple):
+@dataclasses.dataclass(frozen=True)
+class LogPosterior:
+    """
+    Class holding the result of a log-posterior computation, including log-priors,
+    log-likelihoods and derived parameters.
+    """
+
     logpost: float
     logpriors: Sequence[float]
     loglikes: Sequence[float]
     derived: Sequence[float]
+    logprior: float = dataclasses.field(init=False)
+    loglike: float = dataclasses.field(init=False)
+
+    def __post_init__(self):
+        """Sets the sums of priors and posterior."""
+        object.__setattr__(self, 'logprior', sum(self.logpriors))
+        object.__setattr__(self, 'loglike', sum(self.loglikes))
 
 
 class Requirement(NamedTuple):
--- a/cobaya/samplers/evaluate/evaluate.py
+++ b/cobaya/samplers/evaluate/evaluate.py
@@ -69,12 +69,12 @@ class Evaluate(Sampler):
                 logpost=self.logposterior.logpost, logpriors=self.logposterior.logpriors,
                 loglikes=self.logposterior.loglikes)
             self.log.info("log-posterior  = %g", self.logposterior.logpost)
-            self.log.info("log-prior      = %g", sum(self.logposterior.logpriors))
+            self.log.info("log-prior      = %g", self.logposterior.logprior)
             for j, name in enumerate(self.model.prior):
                 self.log.info(
                     "   logprior_" + name + " = %g", self.logposterior.logpriors[j])
-            if sum(self.logposterior.logpriors) > -np.inf:
-                self.log.info("log-likelihood = %g", sum(self.logposterior.loglikes))
+            if self.logposterior.logprior > -np.inf:
+                self.log.info("log-likelihood = %g", self.logposterior.loglike)
                 for j, name in enumerate(self.model.likelihood):
                     self.log.info(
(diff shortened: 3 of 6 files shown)


docs/src_examples/cosmo_model/3.py
```diff
@@ -5,8 +5,9 @@
               'logA': 3.06, 'n_s': 0.966, 'tau_reio': 0.065})
 
 logposterior = model.logposterior(point)
+logposterior_dict = logposterior.as_dict(model)
```
JesusTorrado, line 8: Example of simplification with LogPosterior.as_dict(model)

docs/src_examples/cosmo_model/3.py
```diff
@@ -5,8 +5,9 @@
...
 print('Full log-posterior:')
-print('   logposterior: %g' % logposterior.logpost)
-print('   logpriors: %r' % dict(zip(list(model.prior), logposterior.logpriors)))
-print('   loglikelihoods: %r' % dict(zip(list(model.likelihood), logposterior.loglikes)))
-print('   derived params: %r' % dict(zip(list(model.parameterization.derived_params()), logposterior.derived)))
+print('   logposterior:', logposterior_dict["logpost"])
+print('   logpriors:', logposterior_dict["logpriors"])
+print('   loglikelihoods:', logposterior_dict["loglikes"])
+print('   derived params:', logposterior_dict["derived"])
```
JesusTorrado, line 13: Example of simplification with LogPosterior.as_dict(model)

cobaya/model.py
```diff
@@ -399,39 +528,35 @@ def logposterior(self, params_values, return_derived=True,
...
-            logpriors = [logps]
+            logpriors = [logpriors_1d]
             if self.prior.external:
                 logpriors.extend(self.prior.logps_external(input_params))
-                logpost = sum(logpriors)
-            else:
-                logpost = logps
-
-        if logps != -np.inf:
+        if logpriors_1d != -np.inf:
```
JesusTorrado, line 541: This check (both before and after the PR) goes on to compute the likelihoods even if "external" priors are null. I don't think it should be that way: sometimes poeple use external priors to exclude large parameter regions, and I would prefer not to waste time computing the likelihood there. @cmbant if you think this behaviour is useful, maybe we can make it optional?

JesusTorrado:
@cmbant This should be ready. Do you want to take a look? (esp the code comments above)

cobaya/collection.py
```diff
@@ -178,32 +178,34 @@ def reset(self):
         if getattr(self, "file_name", None):
             self._n_last_out = 0
 
-    def add(self, values, derived=None, weight=1,
-            logpost=None, logpriors=None, loglikes=None):
+    def add(self, values: np.ndarray,
```
cmbant, line 181: values is a list not ndarray in evaluate.py

cmbant (commented):
I didn't look in detail what dataclass is doing but looks OK. Some things are getting a bit hard to understand with so many different permutations of return objects.

cobaya/model.py
cmbant: Probably just an oversight, fine my me to change.

cobaya/model.py
```diff
@@ -476,6 +608,8 @@ def get_valid_point(self, max_tries, ignore_fixed_ref=False, random_state=None):
                                             "or a pdf.")
             raise LoggedError(self.log, "Could not find random point giving finite "
                                         "posterior after %g tries", max_tries)
+        if logposterior_as_dict:
+            results = results.as_dict(logposterior_as_dict)
```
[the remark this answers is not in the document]
JesusTorrado: Oops, sorry. My emacs config should have caught these ones too, ideally, but my ``mypy`` config (the program that does these checks, possibly within pycharm too) is acting up lately.

cobaya/model.py
```diff
@@ -44,12 +45,83 @@ def timing_on(model: 'Model'):
...
+    """
+    Class holding the result of a log-posterior computation, including log-priors,
+    log-likelihoods and derived parameters.
+
+    If ``finite=True`` (default: False), it will try to represent infinities as the
+    largest real numbers allowed by machine precision.
+    """
+
+    logpost: float = None
+    logpriors: Sequence[float] = None
```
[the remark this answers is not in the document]
JesusTorrado: Thanks! I did look this one up at some point, because I want's sure, but I guess I got outdated info. Updated!

JesusTorrado:
> I didn't look in detail what dataclass is doing but looks OK. Some things are getting a bit hard to understand with so many different permutations of return objects.

Thanke for taking a look!

I agree that the typing is clunky when there are multiple possible return values, but I think the code and docstrings are comprehensible when selecting these values (``return_derived``, ``as_dict``). Do you have a better solution in mind that keeps the flexibility? For removing e.g. ``return_derived``, we would need a separate function for derived parameters, and that would involve caching at the model level, which I'd say it's not ideal. As for removing ``as_dict``, it would make it a bit slower, since some functions do (and did) store stuff internally as dictionaries and flattened to lists before returning, so there would be 2 extra conversions if we want dicts as output.

Dataclass is supposed to do pretty much the same as NamedTuple, but with added flexibility. As for speed, see https://stackoverflow.com/questions/51671699/data-classes-vs-typing-namedtuple-primary-use-cases. It states that access time is comparable for small number of attributes (a bit faster for Dataclasses). Creation is a different story: checking for old vs new LogPrior implementation it appears to be ~1mus for NamedTuple vs ~50mus for DataClass. For member access, I can confirm that access time is comparable, and when it comes to getting the summed loglikes/logpriors, the Dataclass that caches them at init is 3 times faster than accessing+summing in the NamedTuple case (and that is done a number of times in the code per call, especially if any "add" method of Collection is called). Anyway, let me see if I can do a bit more profiling before merging, to see where the time difference in creation is going, whether I can do anything about it, and how much does it matter.

JesusTorrado:
Found the culprit: there was an unnecessary ``numpy.isclose`` call. Now the initialisation time is comparable: ~1mus difference, much smaller than the largest overhead call in the best case scenario: the conversion from the sampled to the input parameters. (And this is in the optimistic case in which summed logpriors and loglikes is only requested once, which is not always the case)

image

BTW, backwards compatibility is guaranteed at the syntax level between ``NamedTuple`` and ``dataclass``, both at initialsation and member access. So no worries.

Merging if you agree? Thanks for checking!

cobaya/model.py
```diff
@@ -264,19 +385,38 @@ def logps(self, input_params, return_derived=True, cached=True, make_finite=Fals
...
-        if return_derived:
-            # Turn the derived params dict into a list and return
+        if as_dict:
+            loglikes = dict(zip(self.likelihood, loglikes))
+        if return_derived or return_output_params:
             if not compute_success:
-                derived_list = [np.nan] * len(self.output_params)
+                return_params_names = (self.output_params if return_output_params else
+                                       self.parameterization.derived_params)
+                if as_dict:
```
cmbant, line 394: Should maybe make self.derived_params list similar to self.out_params for consistency and avoid dict copy in parameterization.derived_params?

cmbant:
The only place where speed might be an issue would be in the dragging loop where lots of proposals are rejected quickly by the prior, but even then would agree likely negligible.

cobaya/model.py
JesusTorrado: Done, thanks!

PR #195, opened 2021-08-26
return CAMB lensed_scalar Cls

kimmywu:
This is a small tweak that allows the likelihood to retrieve the lensed scalar Cls in addition to lensed total Cls. I have only written this for CAMB but not for CLASS as I am less familiar with the latter. Happy to get feedback and make improvements. Thanks!


JesusTorrado:
Wouldn't it make the code simpler to just add a `component: "s"|"t"|"st"(or None, default)` to the likelihood requirement and a similar `component` kwarg to `get_lensed|unlensed_cl`? It would save us quite a bit of almost duplicated code and docstrings, and also be similar to what `get_Pk_interpolator` does with its kwarg `var_pair`.

kimmywu:
@JesusTorrado: I like your suggestion. However, if we change the likelihood requirements to ask for `lensed_Cl` in place of `Cl`, it will be a code breaking change. I think a good time for your suggestion to be implemented is when the code could (a)  return both lensed and unlensed scalar Cls, and (b) return these for both CAMB and Class. 

For now, I'd say the current setup adds only minimal extra docstrings and little/no duplicated code. So we add what we have now and do a clean up/code refactor when (a) and (b) above are good to go.

msyriac:
I haven't followed all of this PR and discussion, but thought this might be relevant: is there a way to get the CAMB output of CAMBdata.get_lensed_gradient_cls in Cobaya too? This is useful for e.g. lensing norm likelihood corrections (though in practice, lensed Cls work fine).

cmbant:
You can request the CAMBdata object and get it from that.  Or to separate the calculation from the likelihood, make a separate Theory requesting CAMBdata fron CANV and saving the gradient Cls in the Theory.calculate method.

cmbant:
This seems harmless and potentially useful so have merged for now

kimmywu:
Great. Thank you!

PR #204, opened 2021-10-05
Adding the BICEP/Keck 2021 data release

doicbek:
This PR comprises of several, mostly minor modifications to replace the built-in BICEP/Keck 2015 (BK15) likelihood code with the BK18 likelihood conforming with the latest data release on October 4th 2021. 

**bicep_keck_2018 likelihood**
Besides updating the `download_url`  I exposed `lform_dust_decorr` and `lform_sync_decorr` as likelihood parameters in order to be able to run all likelihood variations presented in the paper.

**bicep_keck_2018 yaml file**
The dataset file was updated. Frequency maps now include 95 GHz, 150 GHz and 220 GHz apodized to the BICEP2 and Keck observing regions and a 95 GHz map apodized to the BICEP3 observing region. WMAP and Planck maps are apodized to the larger BICEP3 map size. Another change compared to BK15 is the replacement of the Gaussian prior on beta_d with a flat prior on range [0.80, 2.40] in the baseline analysis.

The unit test in `test/test_cosmo_bicep_keck_2015.py->test/test_cosmo_bicep_keck_2018.py` was changed accordingly.


cmbant:
Great thanks. Test fail looks like trivial whitespace issue.

You might need to keep alias to BK15 to use map to covariances from the planck chains, or is the result so different that it would be worse than nothing? (would be nice to have a planck+BK18 covmat in the supplied covmats).

Did they decide against hosting the files on github? (or anywhere more reliable than bicepkeck.org...)..

doicbek:
> You might need to keep alias to BK15 to use map to covariances from the planck chains, or is the result so different that it would be worse than nothing? (would be nice to have a planck+BK18 covmat in the supplied covmats).

There are BK18 covmats in `data/bicep_keck_2018/BK18_cosmomc/planck_covmats/` including one combined BK18+Planck 2018 (plik TTTEEE, lowl T, lowl E, lensing).

The latest test still fails because a zip compression issue, while it works for me locally on different machines.

PR #222, opened 2021-12-14
Euclid requisites -- CAMB & CLASS

JesusTorrado:
- [x] Omega_b|cdm|nu_massive(z)
  - [x] CAMB
  - [x] CLASS
  - [x] add test
- [x] extrap_kmin for P(k,z) interpolator (common for CLASS and CAMB)
- [x] `angular_diameter_distance_2`
  - [x]  pooling and requesting values
  - [x] CAMB
  - [x] CLASS
  - [x] add test
- [x] CLASS: sigma8(z) -- restore test
- [x] CLASS: fsigma8(z) -- restore test
- [x] try to incorporate https://github.com/CobayaSampler/cobaya/pull/95
- [x] CLASS: Pk_Weyl -- test
- [x] CLASS: Remove `Pk_max` 1.5 factor
- [ ] CLASS: last neutrino case fails in GUI
- [x] CLASS: update to 3.1.X (devel for now)
- [x] Update CHANGELOG

Other:
- [x] added ``--minimize`` flag to ``cobaya-run``

### DIFF
--- a/cobaya/theories/camb/camb.py
+++ b/cobaya/theories/camb/camb.py
@@ -283,6 +283,7 @@ class CAMB(BoltzmannBase):
         if not self.external_primordial_pk \
                 and set(self.input_params).intersection({'r', 'At'}):
             self.extra_attrs["WantTensors"] = True
+            self.extra_attrs["Accuracy.AccurateBB"] = True
 
     def get_can_support_params(self):
         return self.power_params + self.nonlin_params
@@ -724,8 +725,16 @@ class CAMB(BoltzmannBase):
                     self.log.debug("Setting attributes of CAMBparams: %r",
                                    self.extra_attrs)
                 for attr, value in self.extra_attrs.items():
-                    if hasattr(params, attr):
-                        setattr(params, attr, value)
+                    obj = params
+                    if '.' in attr:
+                        parts = attr.split('.')
+                        for p in parts[:-1]:
+                            obj = getattr(obj, p)
+                        par = parts[-1]
+                    else:
+                        par = attr
+                    if hasattr(obj, par):
+                        setattr(obj, par, value)
                     else:
                         raise LoggedError(
                             self.log,
--- a/cobaya/theories/camb/camb.py
+++ b/cobaya/theories/camb/camb.py
@@ -180,7 +180,7 @@ from cobaya.theories.cosmo import BoltzmannBase
 from cobaya.log import LoggedError, get_logger
 from cobaya.install import download_github_release, check_gcc_version, NotInstalledError
 from cobaya.tools import getfullargspec, get_class_methods, get_properties, load_module, \
-    VersionCheckError, str_to_list, Pool1D, combine_1d
+    VersionCheckError, str_to_list, Pool1D, Pool2D, PoolND, check_2d
 from cobaya.theory import HelperTheory
 from cobaya.typing import InfoDict
 
@@ -190,7 +190,7 @@ class Collector(NamedTuple):
     method: Callable
     args: list = []
     kwargs: dict = {}
-    z_pool: Optional[Pool1D] = None
+    z_pool: Optional[PoolND] = None
 
 
 class CAMBOutputs(NamedTuple):
@@ -352,9 +352,9 @@ class CAMB(BoltzmannBase):
                     k, v["z"], CAMBdata.get_Omega, kwargs={"var": varnames[k]})
             elif k in ("angular_diameter_distance", "comoving_radial_distance"):
                 self.set_collector_with_z_pool(k, v["z"], getattr(CAMBdata, k))
-                self.collectors[k] = Collector(
-                    method=getattr(CAMBdata, k),
-                    kwargs={"z": self._combine_z_for_collector(k, v["z"])})
+            elif k == "angular_diameter_distance_2":
+                self.set_collector_with_z_pool(
+                    k, v["z_pairs"], CAMBdata.angular_diameter_distance2, d=2)
             elif k == "sigma8_z":
                 self.add_to_redshifts(v["z"])
                 self.collectors[k] = Collector(
@@ -484,7 +484,7 @@ class CAMB(BoltzmannBase):
             self.z_pool_for_perturbations.update(z)
         self.extra_args["redshifts"] = np.flip(self.z_pool_for_perturbations.values)
 
-    def set_collector_with_z_pool(self, k, zs, method, args=[], kwargs={}):
+    def set_collector_with_z_pool(self, k, zs, method, args=[], kwargs={}, d=1):
         """
         Creates a collector for a z-dependent quantity, keeping track of the pool of z's.
         """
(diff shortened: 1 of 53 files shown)


cobaya/theories/classy/classy.py
```diff
@@ -144,16 +144,20 @@
 from cobaya.log import LoggedError, get_logger
 from cobaya.install import download_github_release, pip_install, NotInstalledError, \
     check_gcc_version
-from cobaya.tools import load_module, VersionCheckError
+from cobaya.tools import load_module, VersionCheckError, Pool1D, Pool2D, PoolND, \
```
cmbant, line 147: pool2D unused

cobaya/theories/classy/classy.py
```diff
@@ -385,6 +428,9 @@ def calculate(self, state, want_derived=True, **params_values_dict):
                     kwargs[arg_array] = v
                     state[product][i] = method(
                         *self.collectors[product].args, **kwargs)
+            else:
+                raise LoggedError(self.logger, "Variable over which to do an array call "
```
cmbant, line 432: Should be self.log ? (your linter should show  this as undefined)

cobaya/tools.py
```diff
@@ -1069,3 +1069,251 @@ def sort_cosmetic(info):
...
+        Returns a correctly formatted array.
+
+        Internal sorting is enforced, but external is is ignored.
+        """
+
+    @abstract
+    def _update_values(self, values):
+        """Combines given and existing pool. Should assign ``self.values``."""
+
+    def find_indices(self, values):
```
cmbant, line 1170: Is this much slower than the old method for large arrays? [seems to be doing an in-python loop]

cobaya/theories/camb/camb.py
```diff
@@ -803,7 +864,7 @@ def get_path(cls, path):
     def is_installed(cls, **kwargs):
         if not kwargs.get("code", True):
             return True
-        log = get_logger(cls.__name__)
+        log = get_logger(cls.__name__, color=cls._logger_color)
```
cmbant, line 866: wrong number of args

cobaya/theories/classy/classy.py
```diff
@@ -298,10 +311,40 @@ def must_provide(self, **requirements):
...
+        if "z" in kwargs:
+            kwargs = deepcopy(kwargs)
+            kwargs["z"] = z_pool.values
+        elif "z" in args_names:
+            args = deepcopy(args)
+            i_z = args_names.index("z")
+            args = args[:i_z] + [z_pool.values] + args[i_z:]
+        else:
+            raise LoggedError(
+                self.logger,
```
cmbant, line 341: self.log

cmbant:
Could do with some more tests of new code?

cmbant:
Regarding minimize, cf https://cosmocoffee.info/viewtopic.php?f=11&t=3562

cobaya/tools.py
```diff
@@ -1069,3 +1069,251 @@ def sort_cosmetic(info):
...
+
+    If `old_list` given, it is assumed to be a sorted and uniquified array (e.g. the
+    output of this function when passed as first argument).
+
+    Uses `np.unique`, which distinguishes numbers up to machine precision.
+    """
+    new_list = np.atleast_1d(new_list)
+    if old_list is not None:
+        new_list = np.concatenate((old_list, new_list))
+    return np.unique(np.sort(new_list))
```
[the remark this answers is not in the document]
JesusTorrado: Sure. Didn't notice np.unique returns sorted.

cobaya/theories/classy/classy.py
JesusTorrado: Sure, didn't notice that one, even though the linter complains. At some point, I'd like to change all the `log` attributes to `logger`, which is more precise. And maybe the `raise LoggedError` to a method?

cmbant: A method would be nice, but I think complicates the stack dump unless you hack it somehow

cobaya/tools.py
JesusTorrado: The old method would return wrong values for input values with a small positive epsilon added, so not a fair comparison. In any case, it depends on slow is much slower. With this test:

    from cobaya.tools import Pool1D
    import numpy as np
    n = 500
    n_test = 100
    values = np.random.random(n)
    pool = Pool1D(values)
    test_values = np.random.choice(values, n_test) + 1e-15 * np.random.random(n_test)
    %timeit pool.find_indices(test_values)

One can check that the times are:

    find 100 in 500: ~10ms
    find 1000 in 5000: ~50ms
    find 10000 in 50000: ~1.3s

which looks reasonable up to the few thousands, I'd say, given the expected costs of the rest of the pipeline. For the >10k case, we could even make it faster (see note in the `PoolND` code about taking advantage of the pool being sorted). If needed, we can even rewrite the relevant part in cython/numba.

cmbant: I'm sure a few thousand is enough. But e.g. could compare with doing old method first and then only doing the new method if the exact value is not found (or more generally, test a hash index first for an exact match, and only search if not found)

JesusTorrado: Good idea! https://github.com/CobayaSampler/cobaya/pull/222/commits/38454acc1af4cd338c17321238ed890c83508b4f does the fast search first, and the slow one for the elements not found in the first round. As fast as the original one if elements are actually equal, but slower if small perturbations introduced. Still guaranteed correct (or if not, at least not failing silently!). Now I need to do it for 2D.

JesusTorrado: Done for 2D too (still has a python loop, but only over test values. At least 10x faster than the slow and coarse alternative). Added non-cosmo test for 1D and 2D

JesusTorrado:
Almost done here (just a segfault in CLASS needing fixing).

- We are not using this global `k_max` anywhere at the moment, right? Could I remove it? https://github.com/CobayaSampler/cobaya/blob/36a9acc9cbab3e2108105654c69d29aa565afed4/cobaya/theories/cosmo/boltzmannbase.py#L133

- When adding `sigma_R` tests, I've noticed that we return `x, y, A[y,x]`, which is a bit weird. Do you think `sigma_R` is used little enough that we can still change that? (I mean, transpose `A`)

cmbant:
Yes, looks like we are not using k_max. CAMB has "kmax" which can be set to force higher kmax. Sigma_R order doesn't bother me, but could make it y, x, A(y,x) I suppose; not sure if anyone uses it.

JesusTorrado:
OK, I'll change it. Any preference for either `z, R, sigma(z,R)` or `R, z, sigma(R,z)`? I guess the second.

Feel free to review if/whenever you want. As soon as the new CLASS release (3.1.2) is out, I could merge.

cmbant:
I think sigma(z,R)  is the "natural" order from camb. I think I has a quick look at most things last review, anything specific new I should look at?

cmbant:
Do a check for unused imports; minimize argument for run not in docstring.

cmbant:
_get_z_dependent has different arguments in camb and base, should probbaly be consistent.

JesusTorrado:
CLASS has been updated, so this can be merged. Some notes:

- `_get_z_dependent` indeed has different arguments. CAMB has 2 different pools of z values, and the way the code is written right now, CAMB's `_get_z_dependent` is in practice a wrapper over the general `Boltzmann` one that passes the right pool. I could give CAMB's re-implemented method a dummy `pool` argument that is never used if that's making your linter squeak (but it's a private method anyway).
- `get_sigma_R` now returns `z, R, sigma(z,R)`.
- Unused imports removed (I think)
- `minimize` added to `run` docstring

I don't think there is anything left to check.

PR #227, opened 2022-01-18
Add COBAYA_USE_FILE_LOCKING env

cmbant:
To allow turning off portalocker, e.g. workaround for issues like https://github.com/CobayaSampler/cobaya/issues/219

### DIFF
--- a/cobaya/output.py
+++ b/cobaya/output.py
@@ -31,6 +31,17 @@ _kind = "txt"
 _ext = "txt"
 
 
+def use_portalocker():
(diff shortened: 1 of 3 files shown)


JesusTorrado:
Happy for it to be merged whenever you want. Maybe add mention to the new envvar in the troubleshooting section of the docs?

cmbant:
I added something to installation.rst. I notice a lot of general trouble shooting advice is currently under cosmo troubleshooting, maybe we should reorganise at some point.

PR #231, opened 2022-02-11
NaNs in Cl fix for PolyChord

williamjameshandley:
In the initial stages of nested sampling, when it is drawing samples deep in the tails of the posterior distribution one can occasionally get NaN values for the Cls. When these are passed to clik code, this can cause segfaults or other undefined behaviour. The safest thing to do is to catch them and return a likelihood of logzero as an 'unphysical' point, consistent with other situations in the code

Credit to @lukashergt for writing and testing this code. I'm in the process of updating cobaya+polychord, and this will be the first in a sequence of pull requests


williamjameshandley:
@JesusTorrado, I think that (a) I've caught all of the relevant portions of the code and (b) that it needs to be checked and acted upon in each of these cases. In principle this could be pushed back into self.provider.check_nan_Cl(Cl) or something similar to avoid code repetition, although having just had a go at that it doesn't actually reduce the number of lines of code.

williamjameshandley:
The Travis CI timed out -- I'm not sure this is related to this update.

cmbant:
I re-ran the failed one and was OK

cmbant:
Why/from where is it giving NaN at all? Would be better to handle this at a higher level somewhere rather than in each likelihood.

JesusTorrado:
@williamjameshandley and @lukashergt, thanks a lot! I'll come back to this as soon as I am finished with https://github.com/CobayaSampler/cobaya/pull/222 (1-2 working days)

As Antony said, ideally clik would do this. Do you have particular configurations that produce this error, so that we can send them to Karim?

But I think that this makes sense as a provisional solution, as opposed to having to tell people to reduce the size of the prior manually before running PolyChord. Have you checked how large is the overhead? (not significant, I guess)

williamjameshandley:
@cmbant -- these nans are very unusual. They only occur when sampling across the full prior range, and only ~O(10^-6) of the time. They're therefore quite hard to catch in debug mode, and are almost certainly from extreme corner cases. I can move the Cl checks further back, but in my view the Cl calculation is probably returning the 'right thing', and the correct thing to do is to catch it at an 'unphysical parameter point' level i.e. at the loglikelihood.

williamjameshandley:
@JesusTorrado, I'll put in some logging statements in my local code to output the parameter values if/when it catches them and get back to you. 

However, I don't think we would send these to Karim. The problem is that click is undefined for nan inputs. Ideally it could return an error if provided nans (and this wouldn't require specific parameter values), but equally you could say it's the user's responsibility to not hand nonsense to the likelihood code. The 'issue' fundamentally lies with in camb for producing nan Cls, although again, I'm not certain it's wrong to do so for such unreasonable inputs. I think that the correct place to handle this is at the likelihood/modelling level.

JesusTorrado:
In that case, it sounds like CAMB/CLASS (which one?) should detect a failed computation and not even run the likelihood, I think.

cmbant:
For CAMB we could add a Collector "post" for Cl results, and raise and error there if any element of the array is NaN?

JesusTorrado:
Maybe the simplest solution is a check just before the clik call, when 
the array containing Cl's and nuisance is constructed. Works for CAMB 
and CLASS.

JesusTorrado:
Just after this:
https://github.com/CobayaSampler/cobaya/blob/d9458386846eb782430d963821522bf1ab47a7ae/cobaya/likelihoods/base_classes/planck_clik.py#L126

cmbant:
But this is not clik specific?

JesusTorrado:
There are actually 2 problems:
- CAMB-specific: failing silently by producing unphysical results, instead of raising an error.
- clik-specific: segafulting instead of raising a Python error when receiving nan's

Ideally both should be fixed in their respective codes, not their Cobaya interface. Since that's not happening, let's fix both in the respective interfaces. (Even if fixing the first one fixes the second one, avoiding a segfault is worth a small amount of overhead.)

Agreed?

@williamjameshandley @lukashergt have you noticed this happening with CLASS too?

lukashergt:
> @williamjameshandley @lukashergt have you noticed this happening with CLASS too?

Yes, I initially ran into this with Cobaya+PolyChord+CLASS+Planck runs.

cmbant:
@williamjameshandley, if this fix is required for runs to work, that means it must be reproducible, so you could identify the underlying problematic models?

JesusTorrado:
`clik` done in master. Waiting for updates on the `camb.mathutils.chi_squared` cases.

PR #232, opened 2022-02-11
Polychord 1.20.0

williamjameshandley:
This branch aligns cobaya with the latest version of polychord (1.20.0), as well as instantiating some functionality which was present in 1.18.2 (e.g. variable nlive, nprior and maximisation).

### DIFF
--- a/cobaya/samplers/polychord/polychord.py
+++ b/cobaya/samplers/polychord/polychord.py
@@ -82,7 +82,7 @@ class polychord(Sampler):
             if getattr(self, p) is not None:
                 setattr(self, p, NumberWithUnits(
                     getattr(self, p), "d", scale=self.nDims, dtype=int).value)
-        self._quants_nlive_units = ["nprior"]
+        self._quants_nlive_units = ["nprior","nfail"]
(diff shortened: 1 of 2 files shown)


JesusTorrado:
@williamjameshandley  Any chance of a minor release in the next few days, including the `__version__` variable that had not been added yet for the v1.20 tag?

It would be quite useful: with the new PR #237 users would get a warning at runtime that they need to install the new PolyChord version (they will have 1.18.2 installed at that point). Otherwise they'll get a non-descriptive error after they update Cobaya, if they haven't run `cobaya-install`.

williamjameshandley:
Any possibility we could work #233 and #231 into the next cobaya release? I have tested them quite thoroughly locally. In particular cobaya does not work out-of-the-box without #231

williamjameshandley:
I've now created a 1.20.1 release (the version number in `cobaya/samplers/polychord/polychord.py` will need to be updated)

JesusTorrado:
Thanks!

#231 for sure. #233 in the next minor (I need to think about being able to use the external priors as such).

Issue #77, opened 2020-01-30
Unexpected logzero behaviour


williamjameshandley:
As discussed in [this anesthetic issue](https://github.com/williamjameshandley/anesthetic/pull/71#discussion_r372246262), there is some slightly unexpected behaviour within `cobaya.samplers.polychord.polychord.polychord.run`. The return of `logpost` is:
```python
 return max(logposterior + self.logvolume, 0.99 * self.pc_settings.logzero), derived
```
I can see two potential problems with this usage:
1)  `self.logvolume`, could in theory be positive or negative. PolyChord returns `pc_settings.logzero` to indicate an unphysical parameter region, so shouldn't really be increased or decreased
2) Cobaya is effectively re-defining the logzero value to be `0.99 pc_settings.logzero`, i.e. turning the points back into 'only just physical'. This causes issues with codes that assume the `logzero` value is interpreted as unphysical regions. What is the motivation for this modification?

I'd like to understand whether there are some deeper cobaya reasons for this before submitting a PR

JesusTorrado:
Hi @williamjameshandley 

Sorry, I've been quite busy with an application.

The 0.99 factor was there because at some point (I think prior to 1.16) PolyChord was not able to take into account "unphysical" regions of priors, e.g. to guess that in the x,y plane for x,y in [0,1], if we assign -inf to the cases with x<y, the resulting evidence should be 1/2 and not one. SInce that's not the case any more, I will remove the 0.99 factor in devel (2.1, to be released quite soon).

Re the logvolume, as you pointed out in the [anesthetic issue](https://github.com/williamjameshandley/anesthetic/pull/71), it's undoing a scale transformation previously applied, so it returns the correct evidence. Indeed, instead of renormalising the prior **density** to [0,1] using `pc_prior` and then passing the pure likelihood, I am using `pc_prior` to just renormalise with respect to the bounds, and pass the *density*-part of the priors together with the likelihood. This way, we can allow for 1d priors for which we don't have an automatic renormalisation to [0,1].

I hope that clarifies it. Thanks again and sorry for the delay!

williamjameshandley:
> The 0.99 factor was there because at some point (I think prior to 1.16) PolyChord was not able to take into account "unphysical" regions of priors, e.g. to guess that in the x,y plane for x,y in [0,1], if we assign -inf to the cases with x<y, the resulting evidence should be 1/2 and not one.

[...]

PR #233, opened 2022-02-11
Fix and extension to the way that PolyChord treats priors

williamjameshandley:
In contrast to #232 and #231, this is a slightly more complicated/controversial PR.

The way priors are treated in nested sampling is quite subtle, in particular the way that it makes a clear separation between prior and likelihood. This was first covered by [Chen, Feroz & Hobson](https://arxiv.org/abs/1908.04655), and is something I'm exploring with @appetrosyan as part of his [supernest](https://pypi.org/project/supernest/) package, and in a soon-to-be-released paper. This has been discussed partially in #77

In implementing a prior, one can either incorporate it as an additional density term added to the loglikelihood, or as a transformation from the unit Hypercube. MCMC approaches see these as the same, but nested sampling views them quite differently. Both recover the same posterior and evidence, although one will get different KL divergences and different runtimes/accuracy (as a rule of thumb, the more information you can put in the prior, the lower the DKL, the faster the compression from prior to posterior, the more accurate the nested sampling)

This PR implements how I would establish this in cobaya, although discussion would be useful to see if I've missed any subtleties. It:
1. sums the loglikelihoods for the loglikelihood calculation (ignoring the prior density terms)
2. implements priors (uniform and gaussian and beyond) via scipy ppf functions, recording the prior density as before as derived parameters
3. removes the 'volume' correction, which I believe is now redundant.
 
The only thing this therefore fails at is the treatment of the SZ prior for the plik TT|TTTEEE likelihood. In this framework, you have to manually add this in as a *likelihood* rather than a prior density, in the exact same way as you do for priors, e.g.
```yaml
likelihood:
  planck_2018_highl_plik.TTTEEE: null
  planck_2018_highl_plik.SZ: 'lambda ksz_norm, A_sz: stats.norm.logpdf(ksz_norm+1.6*A_sz, loc=9.5, scale=3.0)'
```

It would be great if cobaya-cosmo-generator could do this -- any hints on how I could implement this?

### DIFF
--- a/cobaya/samplers/polychord/polychord.py
+++ b/cobaya/samplers/polychord/polychord.py
@@ -248,7 +248,7 @@ class polychord(Sampler):
             for p in self.model.prior:
                 f_paramnames.write("%s*\t%s\n" % (
                     "logprior" + derived_par_name_separator + p,
-                    r"\pi_\mathrm{" + p.replace("_", r"\ ") + r"}"))
+                    r"\log\pi_\mathrm{" + p.replace("_", r"\ ") + r"}"))
             for p in self.model.likelihood:
                 f_paramnames.write("%s*\t%s\n" % (
                     "loglike" + derived_par_name_separator + p,
--- a/cobaya/samplers/polychord/polychord.py
+++ b/cobaya/samplers/polychord/polychord.py
@@ -225,6 +225,8 @@ class polychord(Sampler):
                 theta[i] = self.model.prior.pdf[i].ppf(xi)
             return theta
 
+        if is_main_process():
+            self.dump_paramnames(self.raw_prefix)
         sync_processes()
         self.mpi_info("Calling PolyChord...")
         self.pc.run_polychord(loglikelihood, self.nDims, self.nDerived, self.pc_settings,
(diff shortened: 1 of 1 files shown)


lukashergt:
This PR is basically an update of #104. 

Would be good to copy the Latex correction from there, i.e. here, this applies to the old line 263 (new line 251). It should be `\log\pi` instead of `\pi`.

lukashergt:
> The only thing this therefore fails at is the treatment of the SZ prior for the plik TT|TTTEEE likelihood. In this framework, you have to manually add this in as a _likelihood_ rather than a prior density, in the exact same way as you do for priors, e.g.
> 
> ```yaml
> likelihood:
>   planck_2018_highl_plik.TTTEEE: null
>   planck_2018_highl_plik.SZ: 'lambda ksz_norm, A_sz: stats.norm.logpdf(ksz_norm+1.6*A_sz, loc=9.5, scale=3.0)'
> ```

<br>

Regarding multidimensional priors it is worth noting that excluding part of the parameter space still works at the prior level. 

For example it still works to restrict one parameter `x` to be greater than another parameter `y` by adding the following to your yaml file:
```yaml
prior:
    exclude_prior: "lambda x, y: np.log(x > y)"
```

But if you want to add some finite `logpdf` value, you will have to add it to the loglikelihood instead of the logprior with this PR.

williamjameshandley:
Many thanks @lukashergt  for pointing out these links. This therefore fixes #101, fixes #103 and closes #104.

PR #235, opened 2022-02-15
added files for the BAO DR12 and DR16 LRG likelihoods

markm42:


cmbant:
Thanks! May be better to delete "rs_fid: 1" lines I think (we should also do that in the other DR16 likes).

Don't worry about the  test fail, that's some annoying pandas/numpy version thing nothing to do with this PR.

cmbant:
I fixed the pandas/numpy conflict on master by upgrading pip in the travis runs.

PR #270, opened 2022-11-10
min python 3.8

cmbant:


cmbant:
Spectacular set of mostly different random build failures, from mpi4py to matplotlib to clik, how is all this still so brittle! I don't immediately have fix for any of them (could caching be messing it up?)

PR #271, opened 2022-11-10
Add CamSpec-NPIPE likelihood

earosenberg:


cmbant:
Apart from trivial spacing failing the tests, looks good to me

cmbant:
It would be nice to add at least one unit test for it though, as for CamSpec 2021

cmbant:
@earosenberg I don't think use_fg_residual_model should the params yaml, since it is not a variable parameter?

earosenberg:
There are several params in params_TT_CamSpec.yaml set to constants - are you suggesting those should get moved elsewhere or removed entirely by editing the Python likelihood?

cmbant:
use_fg_residual_model is not a float, so doesn't make sense to sample; but fine I think to keep things that could be sampled but are fixed by default. Just seems odd having it appearing in outputs yamls like sampleable parameters.

PR #282, opened 2023-02-16
Add lowl TT likelihood ported from clik

eirikgje:
The port is taken from comm_gauss_br_mod_v3.f90. Requires some precomputed files, indicated in the code.

### DIFF
--- a/cobaya/likelihoods/planck_2018_lowl/TT_native.py
+++ b/cobaya/likelihoods/planck_2018_lowl/TT_native.py
@@ -16,7 +16,8 @@ class TT_native(InstallableLikelihood):
                        "github_release": "v1",
                        "asset": "planck_2018_lowT.zip",
                        "directory": "planck_2018_lowT_native"}
-
+    lmin = 2
+    lmax = 29
     type = "CMB"
     aliases = ["lowT"]
 
@@ -26,49 +27,47 @@ class TT_native(InstallableLikelihood):
         return TT.get_bibtex()
 
     def get_requirements(self):
-        return {'Cl': {'tt': self._lmax}}
+        return {'Cl': {'tt': self.lmax}}
 
     def get_can_support_params(self):
         return ['A_planck']
 
-    def initialize(self, lmin=2, lmax=29):
+    def initialize(self):
         if self.get_install_options() and self.packages_path:
-            if lmin < 2 or lmax > 200 or lmin >= lmax:
+            if self.lmin < 2 or self.lmax > 200 or self.lmin >= self.lmax:
                 raise LoggedError(
                     self.log, "lmin must be >= 2, lmax must be <= 200,\n"
                               "and lmin must be less than lmax.")
-            self._lmin = lmin
-            self._lmax = lmax
             #The inverse covariance matrix for the gaussian likelihood calculation
-            self._covinv = np.zeros((self._lmax - self._lmin + 1,
-                                     self._lmax - self._lmin + 1))
+            self._covinv = np.zeros((self.lmax - self.lmin + 1,
+                                     self.lmax - self.lmin + 1))
             # The average cl's for the gaussian likelihood calculation
-            self._mu = np.zeros(self._lmax - self._lmin + 1)
+            self._mu = np.zeros(self.lmax - self.lmin + 1)
             # The cl's used for offset calculation - hence the full range of ells
-            self._mu_sigma = np.zeros(self._lmax + 1)
+            mu_sigma = np.zeros(self.lmax + 1)
 
             # The txt files start at l=2, hence the index gymnastics
             path = self.get_path(self.packages_path)
             self._covinv[:, :] = np.loadtxt(
                 os.path.join(path, 'covinv.txt'))[
-                    self._lmin - 2:self._lmax + 1 - 2,
-                    self._lmin - 2:self._lmax + 1 - 2]
+                    self.lmin - 2:self.lmax + 1 - 2,
+                    self.lmin - 2:self.lmax + 1 - 2]
             self._mu[:] = np.loadtxt(
-                os.path.join(path, 'mu.txt'))[self._lmin - 2:self._lmax + 1 - 2]
-            self._mu_sigma[self._lmin:] = np.loadtxt(
-                os.path.join(path, 'mu_sigma.txt'))[self._lmin - 2:self._lmax + 1 - 2]
+                os.path.join(path, 'mu.txt'))[self.lmin - 2:self.lmax + 1 - 2]
+            mu_sigma[self.lmin:] = np.loadtxt(
+                os.path.join(path, 'mu_sigma.txt'))[self.lmin - 2:self.lmax + 1 - 2]
 
             # Spline info
             nbins = 1000
-            cl2x = np.zeros((nbins, self._lmax - self._lmin + 1, 2))
-            cl2x[:, :, 0] = np.loadtxt(os.path.join(path, 'cl2x_1.txt'))[:, self._lmin - 2:self._lmax + 1 - 2]
-            cl2x[:, :, 1] = np.loadtxt(os.path.join(path, 'cl2x_2.txt'))[:, self._lmin - 2:self._lmax + 1 - 2]
+            cl2x = np.zeros((nbins, self.lmax - self.lmin + 1, 2))
+            cl2x[:, :, 0] = np.loadtxt(os.path.join(path, 'cl2x_1.txt'))[:, self.lmin - 2:self.lmax + 1 - 2]
+            cl2x[:, :, 1] = np.loadtxt(os.path.join(path, 'cl2x_2.txt'))[:, self.lmin - 2:self.lmax + 1 - 2]
             self._spline = []
             self._spline_derivative = []
 
             # Set up prior and spline
-            self._prior_bounds = np.zeros((self._lmax + 1 - self._lmin, 2))
-            for i, l in enumerate(range(self._lmax - self._lmin + 1)):
+            self._prior_bounds = np.zeros((self.lmax + 1 - self.lmin, 2))
+            for l in range(self.lmax - self.lmin + 1):
                 j = 0
                 while abs(cl2x[j, l, 1] + 5) < 1e-4:
                     j += 1
@@ -81,11 +80,11 @@ class TT_native(InstallableLikelihood):
                     InterpolatedUnivariateSpline(cl2x[:, l, 0], cl2x[:, l, 1]))
                 self._spline_derivative.append(self._spline[-1].derivative())
 
-            self._offset = self.log_likelihood(self._mu_sigma, init=True)
+            self._offset = self.log_likelihood(mu_sigma, init=True)
 
 
     def log_likelihood(self, cls_TT, calib=1, init=False):
-        theory = cls_TT[self._lmin:self._lmax + 1] / calib ** 2
+        theory = cls_TT[self.lmin:self.lmax + 1] / calib ** 2
 
         if any(theory < self._prior_bounds[:, 0]) or any(theory > self._prior_bounds[:, 1]):
             return - np.inf
@@ -100,7 +99,7 @@ class TT_native(InstallableLikelihood):
             if dxdCl < 0:
                 return -np.inf
             logl += np.log(dxdCl)
-            delta = x - self._mu
+        delta = x - self._mu
         logl += -0.5 * self._covinv.dot(delta).dot(delta)
 
         if not init:
--- a/cobaya/likelihoods/planck_2018_lowl/TT_native.py
+++ b/cobaya/likelihoods/planck_2018_lowl/TT_native.py
@@ -69,17 +69,17 @@ class TT_native(InstallableLikelihood):
 
             # Set up prior and spline
             self._prior_bounds = np.zeros((self.lmax + 1 - self.lmin, 2))
-            for l in range(self.lmax - self.lmin + 1):
+            for i in range(self.lmax - self.lmin + 1):
                 j = 0
-                while abs(cl2x[j, l, 1] + 5) < 1e-4:
+                while abs(cl2x[j, i, 1] + 5) < 1e-4:
                     j += 1
-                self._prior_bounds[l, 0] = cl2x[j + 2, l, 0]
+                self._prior_bounds[i, 0] = cl2x[j + 2, i, 0]
                 j = nbins - 1
-                while abs(cl2x[j, l, 1] - 5) < 1e-4:
+                while abs(cl2x[j, i, 1] - 5) < 1e-4:
                     j -= 1
-                self._prior_bounds[l, 1] = cl2x[j - 2, l, 0]
+                self._prior_bounds[i, 1] = cl2x[j - 2, i, 0]
                 self._spline.append(
-                    InterpolatedUnivariateSpline(cl2x[:, l, 0], cl2x[:, l, 1]))
+                    InterpolatedUnivariateSpline(cl2x[:, i, 0], cl2x[:, i, 1]))
                 self._spline_derivative.append(self._spline[-1].derivative())
 
             self._offset = self.log_likelihood(mu_sigma, init=True)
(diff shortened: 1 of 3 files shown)


cobaya/likelihoods/planck_2018_lowl/TT_native.py
```diff
@@ -0,0 +1,157 @@
...
+    type = "CMB"
+    aliases = ["lowT"]
+
+    _lmin = 2
+    _lmax = 29 # Could in principle be extended to 200
+    _delta_l = 1000
+    _nbins = 0
+
+
+    data = {}
```
cmbant, line 26: Private variables should start with underscore (things that can't be changed via input yaml options) - i.e. data variables like the ones above.

cobaya/likelihoods/planck_2018_lowl/TT_native.py
```diff
@@ -0,0 +1,157 @@
...
+            if dxdCl < 0:
+                return -1e30
+            else:
+                logl += np.log(dxdCl)
+        if not init:
+            logl -= self.data['offset']
+
+        return logl
+
+    def _splint_gauss_br(self, xa, ya, y2a, x):
```
cmbant, line 110: Use standard scipy interpolate functions instead? (presumably faster and neater)

eirikgje: Yes, I would have liked to do this, it just was a bit more work at the moment because I have to figure out whether the original data files 'match' the scipy interpolate functions, as I cannot recreate the data files. But I suppose I could recreate them using the original splinting functions.

cobaya/likelihoods/planck_2018_lowl/TT_native.py
eirikgje: Should be fixed now

cmbant:
The tests are failing at the moment because of trivial whitespace issues (see links to details above). The easiest thing is to use Pycharm's Auto-format, which should get all the whitespace self-consistent.
You can check it locally using
```
flake8 cobaya --select=E713,E704,E703,E714,E741,E10,E11,E20,E22,E23,E25,E27,E301,E302,E304,E9,F405,F406,F5,F6,F7,F8,W1,W2,W3,W6 --show-source --statistics
```

cmbant:
I think it is also not using the calibration parameter, and the install_options "directory" does not match the naming in the zip file.

cobaya/likelihoods/planck_2018_lowl/TT_native.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    Python translation of the Planck 2018 Gibbs TT likelihood (python Eirik Gjerløw, Feb 2023)
+    See https://wiki.cosmos.esa.int/planck-legacy-archive/index.php/CMB_spectrum_%26_Likelihood_Code
+
+    """
+
+    install_options = {"github_repository": "CobayaSampler/planck_native_data",
+                       "github_release": "v1",
+                       "asset": "planck_2018_lowT.zip",
+                       "directory": "planck_2018_lowT_native"}
```
cmbant, line 15: Doesn't match zip?

cobaya/likelihoods/planck_2018_lowl/TT_native.py
```diff
@@ -0,0 +1,157 @@
...
+            self._data['cl2x'][:, 2:, 2] = np.loadtxt(os.path.join(path, 'cl2x_3.txt')) 
+            self._data['mu'][2:] = np.loadtxt(os.path.join(path, 'mu.txt'))
+            self._data['mu_sigma'][2:] = np.loadtxt(os.path.join(path, 'mu_sigma.txt'))
+            _nbins = len(self._data['cl2x'][:, 0, 0])
+
+            # Bandlimit covariance matrix
+            for l in range(self._lmin, self._lmax+1):
+                for k in range(self._lmin, self._lmax+1):
+                    if abs(l-k) > self._delta_l:
+                        self._data[cov][l, k] = 0
```
cmbant, line 55: cov is undefined

cmbant: If it works, code is probably redundant?

cobaya/likelihoods/planck_2018_lowl/TT_native.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def get_requirements(self):
+        return {'Cl': {'tt': self._lmax}}
+
+
+    def log_likelihood(self, cls_TT, init=False):
+
+        if (any(cls_TT[self._lmin:self._lmax+1] < self._data['prior'][self._lmin:self._lmax+1, 0]) or
+                any(cls_TT[self._lmin:self._lmax+1] > self._data['prior'][self._lmin:self._lmax+1, 1])):
+            return -1e30
```
cmbant, line 82: Use -np.inf for rejection

cobaya/likelihoods/planck_2018_lowl/TT_native.py
```diff
@@ -0,0 +1,157 @@
...
+                                 (x[self._lmin:self._lmax+1] - self._data['mu'][self._lmin:self._lmax+1])))
+
+        # Add Jacobian term
+        for l in range(self._lmin, self._lmax+1):
+            dxdCl = self._splint_deriv_gauss_br(self._data['cl2x'][:, l, 0],
+                                               self._data['cl2x'][:, l, 1],
+                                               self._data['cl2x'][:, l, 2],
+                                               cls_TT[l])
+            if dxdCl < 0:
+                return -1e30
```
cmbant, line 102: -np.inf

cobaya/likelihoods/planck_2018_lowl/TT_native.py
```diff
@@ -0,0 +1,157 @@
...
+    _delta_l = 1000
+    _nbins = 0
+
+
+    _data = {}
+    _data['cov'] = np.zeros((251, 251))
+    _data['cl2x'] = np.zeros((1000, 251, 3))
+    _data['mu'] = np.zeros(251)
+    _data['mu_sigma'] = np.zeros(251)
+
```
cmbant, line 31: Looks like these could just be variables rather than elements of a dictionary (better for code validators)

cmbant: The 2:lmax range could also be set up in initialize rather than having explicit cuts later at likelihood run time.

cmbant: (I'm not  very sure what cl2x is - some comments might be helpful!)

cobaya/likelihoods/planck_2018_lowl/TT_native.py
```diff
@@ -0,0 +1,157 @@
...
+        x = np.zeros(self._lmax+1)
+        for l in range(self._lmin, self._lmax+1):
+            x[l] = self._splint_gauss_br(self._data['cl2x'][:, l, 0],
+                                        self._data['cl2x'][:, l, 1],
+                                        self._data['cl2x'][:, l, 2],
+                                        cls_TT[l])
+        logl = -0.5 * sum((x[self._lmin:self._lmax+1] - self._data['mu'][self._lmin:self._lmax+1]) * 
+                          np.dot(self._data['cov'][self._lmin:self._lmax+1, self._lmin:self._lmax+1],
+                                 (x[self._lmin:self._lmax+1] - self._data['mu'][self._lmin:self._lmax+1])))
+
```
cmbant, line 94: Calculate delta once and use numpy dot for both contractions? (covinv.dot(delta).dot(delta))

cobaya/likelihoods/planck_2018_lowl/TT_native.py
eirikgje: Yeah, I didn't really understand this - so the "directory" is meant to be the directory within the zip file once it has been unzipped?

cobaya/likelihoods/planck_2018_lowl/TT_native.py
eirikgje: I'll remove this part! Indeed it is redundant.

cobaya/likelihoods/planck_2018_lowl/TT_native.py
```diff
@@ -0,0 +1,157 @@
...
+                j = 0
+                while abs(self._data['cl2x'][j, l, 1] + 5) < 1e-4:
+                    j += 1
+                self._data['prior'][l, 0] = self._data['cl2x'][j+2, l, 0]
+                j = _nbins-1
+                while abs(self._data['cl2x'][j, l, 1] - 5) < 1e-4:
+                    j -= 1
+                self._data['prior'][l, 1] = self._data['cl2x'][j-2, l, 0]
+
+            np.savetxt('prior.txt', self._data['prior'])
```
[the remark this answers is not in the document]
eirikgje: Debug purposes.. I'll remove that as well!

cobaya/likelihoods/planck_2018_lowl/TT_native.py
cmbant: Actually maybe that line is OK, more the issue that your zip file has everything in a subdirectory. (I would just compare with e.g. how EE_native and other files are set up.) But if it works already, just ignore me! You can test with "cobaya-install planck_2018_lowl.TT_native"

cmbant:
I think using scipy works OK, and the third file from the zip is redundant. This seems to works to high precision if I put the zip files in the root folder and delete the empty file and cl2x_3.txt.

```
import os
import numpy as np
from cobaya.likelihoods.base_classes import InstallableLikelihood
from scipy.interpolate import InterpolatedUnivariateSpline


class TT_native(InstallableLikelihood):
    """
    Python translation of the Planck 2018 Gibbs TT likelihood (python Eirik Gjerløw, Feb 2023)
    See https://wiki.cosmos.esa.int/planck-legacy-archive/index.php/CMB_spectrum_%26_Likelihood_Code

    """

    install_options = {"github_repository": "CobayaSampler/planck_native_data",
                       "github_release": "v1",
                       "asset": "planck_2018_lowT.zip",
                       "directory": "planck_2018_lowT_native"}

    type = "CMB"
    aliases = ["lowT"]

    _lmin = 2
    _lmax = 29  # Could in principle be extended to 200
    _delta_l = 1000
    _n_cl = 251

    _cov_inv: np.ndarray
    _log_offset: float
    _data_mean: np.ndarray
    _prior_bounds: np.ndarray

    @classmethod
    def get_bibtex(cls):
        from cobaya.likelihoods.planck_2018_lowl.TT import TT
        return TT.get_bibtex()

    def initialize(self):
        if self.get_install_options() and self.packages_path:
            path = self.get_path(self.packages_path)
            covinv = np.loadtxt(os.path.join(path, 'cov.txt'))
            self._cov_inv = covinv[self._lmin - 2:self._lmax + 1 - 2,
                            self._lmin - 2:self._lmax + 1 - 2].copy()
            cl2x = np.zeros((self._delta_l, self._n_cl, 2))
            cl2x[:, 2:, 0] = np.loadtxt(os.path.join(path, 'cl2x_1.txt'))
            cl2x[:, 2:, 1] = np.loadtxt(os.path.join(path, 'cl2x_2.txt'))

            data_mean = np.loadtxt(os.path.join(path, 'mu.txt'))
            self._data_mean = data_mean[self._lmin - 2:self._lmax + 1 - 2]
            mu_sigma = np.zeros(self._n_cl)
            mu_sigma[2:] = np.loadtxt(os.path.join(path, 'mu_sigma.txt'))
            _nbins = len(cl2x[:, 0, 0])

            # Set up prior
            self._prior_bounds = np.empty((self._lmax - self._lmin + 1, 2))
            self.spline = []
            self.spline_derivative = []
            for i, L in enumerate(range(self._lmin, self._lmax + 1)):
                j = 0
                while abs(cl2x[j, L, 1] + 5) < 1e-4:
                    j += 1
                self._prior_bounds[i, 0] = cl2x[j + 2, L, 0]
                j = _nbins - 1
                while abs(cl2x[j, L, 1] - 5) < 1e-4:
                    j -= 1
                self._prior_bounds[i, 1] = cl2x[j - 2, L, 0]

                self.spline.append(
                    InterpolatedUnivariateSpline(cl2x[:, L, 0], cl2x[:, L, 1]))
                self.spline_derivative.append(self.spline[-1].derivative())

            self._log_offset = 0
            self._log_offset = self.log_likelihood(mu_sigma)

    def get_requirements(self):
        return {'Cl': {'tt': self._lmax}}

    def get_can_support_params(self):
        return ['A_planck']

    def log_likelihood(self, cls_TT, calib=1):
        theory = cls_TT[self._lmin:self._lmax + 1] / calib ** 2

        if (any(theory < self._prior_bounds[:, 0]) or
                any(theory > self._prior_bounds[:, 1])):
            return -np.inf

        # Convert the cl's to Gaussianized variables and get jacobian term
        x = np.zeros_like(theory)
        logl = 0
        for i, (spline, diff_spline, cl) in enumerate(
                zip(self.spline, self.spline_derivative, theory)):
            x[i] = spline(cl)
            dxdCl = diff_spline(cl)
            if dxdCl < 0:
                return -np.inf
            logl += np.log(dxdCl)

        # get delta and add chi2 term to log like
        x -= self._data_mean
        logl += -0.5 * self._cov_inv.dot(x).dot(x)

        logl -= self._log_offset

        return logl

    def logp(self, **params_values):
        cls = self.provider.get_Cl(ell_factor=True)['tt']
        return self.log_likelihood(cls, params_values.get('A_planck', 1))


```

eirikgje:
Wow, okay, good job there. I've now changed the other things you've commented; I'll add another commit with this interpolation instead then.

cmbant:
Great thanks. I updated the zip file now, so hopefully sporting chance unit test may work if you add one.

eirikgje:
Could I ask for another change to the zipfile, just for cosmetic purposes? Rename cov.txt -> covinv.txt?

cmbant:
Sure. Could also make it much smaller (and faster load/test) by pre-cutting the matrices, but depends on whether you want to be able to increase lmax.

If you do want people to be able to change lmax, perhaps should make _lmax and _lmin public variables (with initialization test that they are in range).

cmbant:
I think need to add you TT_native to install_tests["likelihood"] in input_database.py for the likelihood to install and then test to actually run (currently marked as XFAIL because not installed).

cmbant:
I think you need to define the A_planck input parameter - either as optionally supported as in my example, or via a yaml file or params attribute.

eirikgje:
It seems the test fails, but it's close.. I don't know how exactly to interpret this

cmbant:
I was seeing agreement at 1e-13. Are your cl2x indices correct, e.g. cf my version?

eirikgje:
Hmm, I'm not sure exactly what you mean. I tested the scipy interpolate solution locally and it gave results that were similar to what I had before, and the cl2x coefficients should be exactly the same as what is in the zip file. But I haven't locally compared it with the non-native lowl TT results, so I can look into how to do that.

cmbant:
To clarify, the 1e-13 was wrt your original python version before and after the change to using scipy, I didn't compare to the clik version at any point. Sorry, I think I didn't notice you'd also changed the definition of "l" in the loop, so it does look consistent also to 1e-13. [you don't actually need "enumerate", because i and l are now the same?] I guess there must be some real difference with the fortran original.

Btw, I think lmin and lmax should be class attributes - cobaya does not pass yaml settings to initialize() optional arguments, but will recognise public attributes.

cmbant:
"delta" should also be defined after the loop, not in it

cmbant:
The difference could be that the inverse of the cut covariance is not the same as the cut of the full inverse covariance (which is what is in python at the mo). I had assumed that cutting the inverse was intentional, on the grounds that you want to likelihood conditional on fixed high-L contrained by another likelihood. But the fortran code seems to be cutting the covariance.

eirikgje:
So, to go a bit deeper in this, I actually wrote a paper about this in the Planck era: https://arxiv.org/pdf/1304.0315.pdf and *ideally* we would evaluate the low-ell likelihood from l=2 to say, l=40, and then evaluate the high-ell likelihood from l=29 to l=whatever, add those two together and then subtract for example the low-ell likelihood evaluated again from l=29 to l=40. I don't think this is what Planck did, so if the goal is fidelity to Planck results, then I'll cut the matrix and invert it after.

cmbant:
Great, many thanks for this - will be very handy to be able to use this without clik (and much more understandable what it is doing!). I made a few trivial refactors in https://github.com/CobayaSampler/cobaya/commit/91cd720403ad6f734d5a08198809f637e2db7b18, hope that's OK.

PR #283, opened 2023-02-22
NPIPE option in input database

cmbant:
We can now run clik-free Planck chains using native lowl TT and EE, +CAMspec NPipe and native lensing. I added a basic config to the input database. 

But the Planck NPIPE lensing likelihood is a separate package at https://github.com/carronj/planck_PR4_lensing/. Would be good to use this in the default NPIPE configuration, but then need some way for install process to know where to get the download from (not currently on pip). Perhaps a keyword "exernal: pip" or "external: github_URL" or similar?


cmbant:
I tried adding package_install option (just starting with github):
https://github.com/CobayaSampler/cobaya/pull/283/files#diff-0417ef553bc88ecdc3d41eea614cc3a85d98bc69c7f8654a3a364bbfbb1dc8d8R141

However this behaves weirdly and I don't understand: 

```
likelihood:
  planckpr4lensing: 
      package_install:
        github_repository: carronj/planck_PR4_lensing
        github_release: master
```
- when not installed, it correctly downloads the code, however when it then tries to load the module it fails. However, it works fine if you then re-run cobaya-install. The module load command appears identical in both cases, but something must be hanging over from the first fail.

cmbant:
@JesusTorrado I think this is basically working with package_install now. There's an example in test_cosmo_planck_NPIPE.py

cmbant:
The mcmc tests are still randomly failing occasionally, but not connected to this PR I think.

Things to consider:

- Replace lowl.TT, lowl.EE with the native ones by default? (moving to TT_clik, or even deleting)
- Thin out the default travis tests so they run quicker/fewer clik installs?
- Longer term it might be nice to strip all the clik and old planck stuff into a separate planck likelihood package, to declutter cobaya.

cmbant:
@JesusTorrado, please also can you check if the gui still works OK on your machine, I had to make some pyside6 changes.

CHANGELOG.md
```diff
@@ -15,11 +16,14 @@
 - Added CAMBspec NPIPE Planck 2020 likelihood (#271) ) (thanks @earosenberg)
 - Added native version of `planck_2018_lowl.EE`.
 - Added native version of `planck_2018_low.TT`. (thanks @eirikgje)
+- Replaced default planck_2018_lowl.EE and planck_2018_low.TT with native versions
```
JesusTorrado, line 19: add that clik is now github

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### General
```
JesusTorrado, line 3: add tempered

PR #288, opened 2023-04-03
Fix a bug in mcmc oversampling

JiangJQ2000:
`indices_repeated` should contain `j`, which was incorrectly written as `i`.

Fortunately, in many simple cases, because the nuisance parameters are appended to the end of the parameter list, oversampling can still work in the presence of bugs. But this is not the case for complex cases.


cmbant:
Thanks, looks like a nasty bug!

I think part of the issue was that the code was too hard to understand, I've tried to tidy it up (with your fix) and cycle blocks more directly in this test branch
https://github.com/CobayaSampler/cobaya/compare/master...test_proposeclean

 @JesusTorrado should probably also check as he did most of the work on oversampling.

PR #293, opened 2023-04-17
add EE_sroll2

cmbant:


cmbant:
The text data file is pulled from the fits file distributed by @paganol as part of a clik bundle. Uses same code as 2018 native EE. Any checks/confirmation of correctness would be helpful.

paganol:
Hi @cmbant, I tested the spectrum and the check_value in the clik package, and the log likelihood matches perfectly. This should be enough. 

```
import numpy as np
from cobaya.likelihoods.planck_2018_lowl.EE_sroll2 import EE_sroll2
from astropy.io import fits

compute_like = EE_sroll2()

h = fits.open('clik/simall_100x143_sroll2_v3_EE_Aplanck.clik/clik/check_param')
l = np.arange(30)
ll = l*(l+1)/2/np.pi
like = compute_like.log_likelihood(h[0].data[0:30]*ll)
np.allclose(like,-195.112851,rtol=1e-8)
```

PR #299, opened 2023-07-10
Scatterwarn

cmbant:


cobaya/samplers/mcmc/mcmc.py
```diff
@@ -146,7 +146,7 @@ def initialize(self):
         if self.output.is_resuming() and len(self.collection):
             last = len(self.collection) - 1
             initial_point = (self.collection[self.collection.sampled_params]
-                             .iloc[last]).to_numpy(dtype=np.float64, copy=True)
+            .iloc[last]).to_numpy(dtype=np.float64, copy=True)
```
JesusTorrado, line 149: pycharm as very weird indentation conventions sometimes...

docs/likelihood_planck.rst
```diff
@@ -1,32 +1,40 @@
...
 - ``planck_2018_highl_CamSpec.[TT|TTTEEE][_native]``: ``clik`` and native Python versions of the alternative high-:math:`\ell` ``CamSpec`` likelihoods.
-- ``planck_2018_highl_CamSpec2021.[TT|TTTEEE]``: native Python versions of updated high-:math:`\ell` ``CamSpec`` likelihoods (following `arXiv 1910.00483 <https://arxiv.org/abs/1910.00483>`_).
 - ``planck_2018_lensing.clik``: lensing temperature+polarisation-based; official ``clik`` code.
 - ``planck_2018_lensing.native``: lensing temperature+polarisation-based; native Python version (more customizable)
 - ``planck_2018_lensing.CMBMarged``: CMB-marginalized, temperature+polarisation-based lensing likelihood; native Python version (more customizable). Do not combine with any of the ones above!
 
+Other more recent Planck likelihoods are:
+
+- ``planck_NPIPE_highl_CamSpec.[TT|TE|EE|TTEE|TTTEEE]``: latest native python NPIPE (PR4) CamSpec high-:math:`\ell` likelihoods
+-  ``planckpr4lensing.[PlanckPR4Lensing|PlanckPR4LensingMarged]``: NPIPE lensing; pip install from `GitHub <https://github.com/carronj/planck_PR4_lensing>`_
```
JesusTorrado, line 29: extra space

docs/likelihood_planck.rst
```diff
@@ -1,32 +1,40 @@
...
 
    <br />
 
-The Planck 2018 likelihoods defined here are:
+The Planck 2018 baseline likelihoods defined here are:
 
-- ``planck_2018_lowl.TT``: low-:math:`\ell` temperature
-- ``planck_2018_lowl.EE``: low-:math:`\ell` EE polarization
+- ``planck_2018_lowl.TT``: low-:math:`\ell` temperature (native python implementation)
+- ``planck_2018_lowl.EE``: low-:math:`\ell` EE polarization (native python implementation)
```
JesusTorrado, line 16: We should mention here the _clik versions of the low-ell ones

docs/installation_cosmo.rst
```diff
@@ -8,12 +8,31 @@ To keep it light, maintainable and easily extensible, **cobaya** does not includ
...
 -----------------------------------------------
 
-To install a comprehensive set of cosmology requisites (CAMB, CLASS, Planck, BICEP-Keck, BAO, SN), in a ``/path/to/packages`` folders of your choice:
+To install a basic set of cosmology requisites (CAMB, CLASS, Planck, BAO, SN), in a ``/path/to/packages`` folders of your choice:
 
 .. code:: bash
 
    $ cobaya-install cosmo -p /path/to/packages
 
+This does not install the Planck 2018 **clik** likelihoods, the original packages can then be installed using e.g.
```
JesusTorrado, line 17: Regarding this block, I think it is not a good approach not to install by default at least one "full" high-ell TTTEEE likelihood. If we do not want clik by default, given how problematic it is, I suggest we add the NPIPE likelihoods to the basic install. The way it is now, `cobaya-install cosmo` is insufficient to run full Planck (PR3 or PR4).

cmbant: I did wonder about that. Can add planck_NPIPE_highl_CamSpec.TTTEEE to install_basic. But the NPIPE lensing is a separate package, so would still not have consistent NPIPE setup and still need to separately run cobaya-install on the actual yaml being used to get the lensing installed. (moving forward I was thinking could encourage people to run cobaya-install on actual yaml, then more things can be external packages)

An alternative would be to add native plik_lite.

PR #301, opened 2023-07-12
Allow sigma8 as input for CAMB

tilmantroester:
This PR allows specifying sigma8 as the normalisation of the power spectrum when using the CAMB theory code. The use case is large-scale structure analyses that sample in sigma8 instead of As. An example is the KiDS-1000 analysis, which samples in S8 to avoid implicit priors on that parameter.

For now this only works with power law initial power spectra, where As can be rescaled and power spectra recomputed after the transfer functions have been computed.

The current implementation seems to do what it's supposed to but I'm not very familiar with cobaya, so if there's a better approach to achieve this, I'd be happy to adapt the implementation.

### DIFF
--- a/cobaya/theories/camb/camb.py
+++ b/cobaya/theories/camb/camb.py
@@ -338,6 +338,14 @@ class CAMB(BoltzmannBase):
                 and set(self.input_params).intersection({'r', 'At'}):
             self.extra_attrs["WantTensors"] = True
             self.extra_attrs["Accuracy.AccurateBB"] = True
+        if "sigma8" in self.input_params and "As" in self.input_params:
+            raise LoggedError(self.log,
+                              "Both As and sigma8 have been provided as input. "
+                              "This will likely cause ill-defined outputs.")
+        if "sigma8" in self.input_params and not self.external_primordial_pk:
+            self.extra_attrs["WantTransfer"] = True
+            self.needs_perts = True
+            self.add_to_redshifts([0.])
 
     def get_can_support_params(self):
         params = self.power_params + self.nonlin_params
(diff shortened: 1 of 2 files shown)


cmbant:
Thanks, interesting, I'll have to think about it. (It's obviously rather inefficient to calculate everything from time transfer functions twice.)

cmbant:
I can't think of a better way of doing it, and looks fine to me and useful for checking (though I would not encourage use given extra numerical cost and arbitrariness of sigma8!). Does it raise an error if someone provides both As and sigma8 as input?

You might be able to do the initial power law check somewhere in initialization instead of once running (initialize_with_params?).

tilmantroester:
I've added a check that raises an error if both As and sigma8 are inputs. I also moved what was in `set` to `initialise_with_params`.

I agree that there are good reasons for not using sigma8 as a parametrisation of the power spectrum amplitude but for better or worse, that's what we're stuck with for low-redshift stuff...

cmbant:
Thanks for the PR

cmbant:
I made a few tweaks, e.g. moving power law check out of run loop and checking also for As. please check
https://github.com/CobayaSampler/cobaya/commit/4a52113b1a3084f8f35a3945931df1fd2a221a5f

tilmantroester:
Looks reasonable to me. Is the reason for using `initialize_with_provider` that `needs_perts` might get changed after `initialize_with_params`?
Thanks for reminding me that the `:=` exists, btw!

cmbant:
Thanks. Yes, it only knows what it's been asked to calculate after all dependencies resolved, which is after initialize_with_params; initialize_with_provider is the last step before actually running a calculation

Issue #202, opened 2021-09-22
Add sample metadata to store information not already in `updated.yaml`


JesusTorrado:
Information that could be stored (in general, stuff about the chains themselves that is not already in the `updated.yaml`):
- whether burn-in has been removed
- whether the sample has been thinned
- whether the sample has been cooled down
- the sample temperature? (I guess not, since it's actually part of the sampler `updated.yaml` block?)

How to store that? A few alternatives:
- Another block (`analysis`?) in the `updated.yaml`
- Another file (could be GetDist's `properties.ini`)

cmbant:
Related to this is the annoying code duplication between getdist and cobaya, e.g. to get the ranges have to parse the parameter blocks. Potentially this could be part of metadata too.

cmbant:
Note grid scripts [326](https://github.com/CobayaSampler/cobaya/pull/326) currently use .properties.ini to store burn-removed info a la getdist.

Issue #166, opened 2021-04-20
Getting MPI output/combined output


cmbant:
We don't have a very clear way to get combined output, and currently some samplers produce the same result on all processes (minimize), and some produce per-process result on each process (post/mcmc). Either add combined_products() to sampler base, or "combined" argument to products() or similar?

JesusTorrado:
FIxed by #315

PR #315, opened 2023-09-11
[WIP] Update collections IO

JesusTorrado:
This branch will address #166 and #202 (only removing code duplication in GetDist)

@cmbant we currently use two different kwargs for combining/concatenating samples in two different places: "combine" in `Sampler.products()` and related ones, and `concatenate` in `Output.load_collections`. In general we are combining, but in practice we are concatenating. Thoughts?

**Update**
- [x] Split from `Output` a non-MPI read-only `OutputReadOnly` to track names. Needed for loading samples in an MPI-independent way while being able to use `Output` goodies (e.g. collection regexps)
- [x] Created `load_samples` functions, importable as `from cobaya import load_samples`, to load samples given a prefix, including skipping, thinning, combining and converting to getdist (see below)
- [x] Added/unified `.products()` and `.samples()` methods for MCMC, PolyChord and post-processing, with options to skip, combine and conver to getdist
- [x] Updated test and docs to use `load_samples` or the two methods above instead of importing `loadMCSamples` or `MCSamplesFromCobaya` from getdist.
- [x] Improve getdist ranges as discussed to take samples into account.
- [ ] Write a getdist PR using `load_samples` so that duplicated code can be dropped from getdist (needs some checks before).


cmbant:
Sounds good. I think you'd have to make a new MCSamples instance passing in the list of samples from each MCSamples

cmbant:
Btw, any reason why are ranges in getdist being set from the 5 sigma confidence interval for unbounded, rather than e.g. well outside the extremal samples?

JesusTorrado:
> Sounds good. I think you'd have to make a new MCSamples instance passing in the list of samples from each MCSamples

I guess I was not doing it wrong then. I was just expecting the final object to be aware of the chains. but it turn out that `MCSamples.makeSingle()` is unconditionally called when there are multiple chains: https://github.com/cmbant/getdist/blob/master/getdist/mcsamples.py#L492C30-L492C30

> Btw, any reason why are ranges in getdist being set from the 5 sigma confidence interval for unbounded, rather than e.g. well outside the extremal samples?

If there aren't extreme-enough samples (quite likely if by extremal we mean 5 sigma), this is "well outside the extremal samples", right? Maybe ranges can be the maximum between 5 sigma and the value of the extremal sample? In any case, I think it is nice to have them be defined in terms of the prior, not the sample, since they are prior ranges. Or maybe I am not fully getting what you mean?

cmbant:
It does combine them, but it also preserves the chain_offsets information so the split information is not lost (e.g. can call getSeparateChains to get list of separate WeightedSamples).

I agree max(5 sigma, extrema difference * 1.1 )) or something might be more appropriate. But with N-> infty samples, setting range at just 5 sigma will guarantee the range is too small to reliably estimate high-sigma contours (obviously this is usually academic in practice..)

JesusTorrado:
@cmbant this is ready for a preliminary review, if you want. The only thing missing before merging is the thing mentioned in your last comment.

cmbant:
Looks like lots of potentially useful things. Have not tried it, just a few notes by eye/pycharm:

I would be inclined not to always instantiate getdist chains with a [samples] rather than samples when there are not multiple chains. e.g. could use 
`samples = samples if combine_with else samples[0]`
etc. or equivalent refactor.

Run now returns PostResult rather than PostResultDict? Also run is tagged to return Sampler, but Sampler.products() is defined as dict. The type checker shows up some incorrect call arguments e.g. in test_cosmo_run.

There are some unused imports like get_literal_param_ranges in model.py.

get_update_info is used in deprecation block but not defined.

JesusTorrado:
Implemented your suggestion about bounds, together with your other comments:

> I would be inclined not to always instantiate getdist chains with a [samples] rather than samples when there are not multiple chains. e.g. could use `samples = samples if combine_with else samples[0]` etc. or equivalent refactor.

Not implemented for now, since it makes this line fail: https://github.com/CobayaSampler/cobaya/blob/4db6c17ca6fdba6e69a0760289f0893c70e8b73d/tests/test_cosmo_run.py#L126

> Run now returns PostResult rather than PostResultDict?

Yes, since an extended object is needed if we want `.samples()` and `.products()` methods compatible with the rest of the samplers, which is nice, IMO. I have implemented a `.__getitem__()` that makes it backwards compatible with `PostResultDict`.

> Also run is tagged to return Sampler, but Sampler.products() is defined as dict.

`run` returns indeed `Sampler`, not `Sampler.products()` any more. But I had forgot to update for the new `PostResult`, which is already in the last commit.

> The type checker shows up some incorrect call arguments e.g. in test_cosmo_run.

I have type-checking disabled in `tests/`, or I'll go crazy, since my type-checker (the standard `mypy`) is very stringent, I guess much more so than that of pycharm. I saw some errors about `Sampler` not having a general/abstract `samples` method and other stuff, which I have tried to fix in the last commit.

> There are some unused imports like get_literal_param_ranges in model.py.
> get_update_info is used in deprecation block but not defined.

Oops, thanks! My linter saw them, but I am working too fast at the moment.

The branch should be feature-complete now (could be improved by adding `.samples()` to `evaluate`, for example, but I need to move to the precision problems.

Waiting on your review or your go-ahead!

cobaya/post.py
```diff
@@ -38,6 +41,113 @@ class PostResultDict(TypedDict):
...
+            The post-processed samples.
+        """
+        # Difference with MCMC: self.results["sample"] may contain one collection or a
+        # list of them pre-process
+        collections = self.results["sample"]
+        if not isinstance(collections, list):
+            collections = [collections]
+        collections = [c.skip_samples(skip_samples, inplace=False) for c in collections]
+        if not (to_getdist or combined):
+            return collections
```
cmbant, line 89: List not consistent with return type

cobaya/samplers/polychord/polychord.py
```diff
@@ -325,107 +333,205 @@ def process_raw_output(self):
...
+        for i, c in self.clusters.items():
             if to_getdist:
-                collection = self.collection.to_getdist(model=self.model)
+                try:
+                    clusters[i] = c["sample"].to_getdist()
+                except (ValueError, AttributeError):
+                    self.log.warning(
+                        "Cluster #%d could not be converted to a GetDist sample. "
+                        "Storing 'None'.", i)
+                    clusters[i] = None
```
cmbant, line 476: None not in declared types

cobaya/post.py
```diff
@@ -38,6 +41,113 @@ class PostResultDict(TypedDict):
     weights: Union[np.ndarray, List[np.ndarray]]
 
 
+class PostResult():
```
cmbant, line 44: Can delete brackets

cmbant: (also OutputReadOnly)

cobaya/samplers/mcmc/mcmc.py
```diff
@@ -961,10 +950,8 @@ def products(
...
             applied previously to concatenation. Forces the return of a copy.
         to_getdist: bool, default: False
-            If ``True``, returns sample collections as :class:'getdist.MCSamples`. If both
-            this option and ``combined`` are ``True``, the latter is ignored and a
-            multi-chain :class:'getdist.MCSamples` object is created out of the chains of
-            all MPI processes.
+            If ``True``, returns a single :class:`getdist.MCSamples` instance, containing
+            all samples (``combined`` is ignored).
 
         Returns
```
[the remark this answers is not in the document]
JesusTorrado: Done. Not a fan of creating a type for something that simple. It is well documented in the Return section

cobaya/samplers/polychord/polychord.py
```diff
@@ -325,107 +333,205 @@ def process_raw_output(self):
...
-                        c["sample"] = c["sample"].to_getdist(model=self.model)
-                products.update({"clusters": clusters})
+                products["clusters"] = {i: {} for i in self.clusters}
+                for i, s in self.samples_clusters(to_getdist=to_getdist).items():
+                    products["clusters"][i]["logZ"] = self.clusters[i]["logZ"]
+                    products["clusters"][i]["logZstd"] = self.clusters[i]["logZstd"]
+                    products["clusters"][i]["sample"] = s
+        do_bcast = combined or to_getdist
+        if do_bcast:
+            return share_mpi(products)
```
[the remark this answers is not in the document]
JesusTorrado: Done, thanks!

cobaya/sampler.py
```diff
@@ -202,9 +207,15 @@ def run(self):
...
         """
-        pass
 
-    def products(self) -> InfoDict:
+    def samples(self, **kwargs) -> Union["SampleCollection", "MCSamples"]:
+        """
+        Returns the products expected in a scripted call of cobaya,
+        (e.g. a collection of samples or a list of them).
+        """
+        return {}
```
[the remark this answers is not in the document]
JesusTorrado: Removed return statements for abstract definitions.

cobaya/post.py
JesusTorrado: Solved, thanks!

cobaya/samplers/polychord/polychord.py
JesusTorrado: Added to type declaration, thanks!

cobaya/post.py
JesusTorrado: Done, thanks!

cmbant:
Great, thanks for all this.
Changing this line in test_cosmo_run to directly get samples clears up type warnings there
`    samp_thin = sampler_init.samples(skip_samples=0.2, to_getdist=True)["sample"]
`

PostResultDict type can contain MCSamples as well (e.g. from post.products).

JesusTorrado:
> Great, thanks for all this. Changing this line in test_cosmo_run to directly get samples clears up type warnings there ` samp_thin = sampler_init.samples(skip_samples=0.2, to_getdist=True)["sample"]`
> 
> PostResultDict type can contain MCSamples as well (e.g. from post.products).

Done, thanks for the review. Ok to merge? I wouldn't squash bc I manged to keep the commits (some at least) quite self contained.

cmbant:
Looks good to me (haven't actually tried running anything)

JesusTorrado:
Ok, merging then. Please let me know if any of the new functionality behaves unexpectedly, whenever you have time to switch to it. I will look at precision stuff now. Thanks for the review!

PR #318, opened 2023-10-06
tidy types etc

cmbant:


cobaya/model.py
```diff
@@ -1354,7 +1354,7 @@ def get_model(info_or_yaml_or_file: Union[InputDict, str, os.PathLike],
     # Inform about ignored info keys
     ignored_info = []
     for k in list(info):
-        if k not in {"params", "likelihood", "prior", "theory", packages_path_input,
+        if k not in {"params", "likelihood", "prior", "theory", "packages_path",
```
JesusTorrado, line 1357: I don't understand the reason for dropping imported names from `conventions`. We should be gong in the opposite direction: hard-coding these in a single place. PyCharm may complain less (if that's the reason), but it is not good programming practice IMO.

cmbant: We don't have that for any of the other key strings, so just seemed more consistent (and makes the TypedDict checking easier, and easier for debugging/reading because you don't have to look up what  "packages_path_input" actually is to see how it relates to the input yaml). We can't actually change the value anyway, since now in the public API for input files... Seem reasonable?

JesusTorrado: We did have that for the other ones too, but you also removed them at some point in practically all modules (with the exception, I think, of field names of single parameter specification), I guess for the same reason, so it's not a consistency problem: I would gladly import them all and make it consistent again.

Even if it did make debugging a bit harder, it does also certainly make easier to write correct code if you import a name than if you try to remember what that name was, so one for the other? Whether we can change them or not at some point, I think it's not so relevant for deciding this.

If I have time at some point, I would like to restore all of these imports in a PR, and then we can try alternatives to solve your type-checking problems in that PR before merging. Would that sound OK?

cmbant: If you really really want to, but doesn't seem very useful. (packages_path_input was particularly annoying because it didn't match the name of the string; over the years I must have clicked on it many times to check that it was actually packages_path and not something else)

I don't see really how remembering a variable name is easier than remembering a string (you have to remember the string when writing yaml anyway). Code completion should be able give the available strings for a typed dict just as well as importing (though not sure which/any IDEs actually do that at the moment, it would be an obvious thing to do...), and at least only editing in one place rather than having to delete/add import statements any time you use/delete a new string.

JesusTorrado: Happy to give the strings and the variables the same name. The "_input" part was there at some point to distinguish between the CLI argument and the name of the field, but I agree that the tradeoff in that case may not be worth it.

It's not a matter of remembering, but also that importing from the same place emphasizes that different parts of the code refer to the same thing, not to a different thing called the same.

Issue #306, opened 2023-08-01
*ERROR* Error when loading samples: The sum of logpriors in the sample is not consistent.


mishakb:
I have this error when I try to resume a job. I was able to resume it at least one time but this second tie it gives this. I tried several times but with same message.  His is the job.out file content: 

```[0 : output] Output to be read-from/written-into folder 'results', with prefix 'ow0waCDM_all'
[0 : output] Found existing info files with the requested output prefix: 'results/ow0waCDM_all'
[0 : output] Let's try to resume/load.
[2 : jax._src.xla_bridge] Unable to initialize backend 'cuda': module 'jaxlib.xla_extension' has no attribute 'GpuAllocatorConfig'
[2 : jax._src.xla_bridge] Unable to initialize backend 'rocm': module 'jaxlib.xla_extension' has no attribute 'GpuAllocatorConfig'
[2 : jax._src.xla_bridge] Unable to initialize backend 'tpu': INVALID_ARGUMENT: TpuPlatform is not available.
[2 : jax._src.xla_bridge] *WARNING* No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)
[0 : jax._src.xla_bridge] Unable to initialize backend 'cuda': module 'jaxlib.xla_extension' has no attribute 'GpuAllocatorConfig'
[0 : jax._src.xla_bridge] Unable to initialize backend 'rocm': module 'jaxlib.xla_extension' has no attribute 'GpuAllocatorConfig'
[0 : jax._src.xla_bridge] Unable to initialize backend 'tpu': INVALID_ARGUMENT: TpuPlatform is not available.
[0 : jax._src.xla_bridge] *WARNING* No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)
[1 : jax._src.xla_bridge] Unable to initialize backend 'cuda': module 'jaxlib.xla_extension' has no attribute 'GpuAllocatorConfig'
[1 : jax._src.xla_bridge] Unable to initialize backend 'rocm': module 'jaxlib.xla_extension' has no attribute 'GpuAllocatorConfig'
[1 : jax._src.xla_bridge] Unable to initialize backend 'tpu': INVALID_ARGUMENT: TpuPlatform is not available.
[...]

PR #322, opened 2023-10-09
post: bugfix: chains should be detempered/reweighted at once (or bad weights)

JesusTorrado:
Should work.

@cmbant  any feedback on the general approach before I go on with the TODO's?

TODO: do the same with `Collection` methods `reweght`, `mean` and `cov`

### DIFF
--- a/cobaya/collection.py
+++ b/cobaya/collection.py
@@ -487,7 +487,8 @@ class SampleCollection(BaseCollection):
                         np.atleast_2d(
                             self[self.minuslogprior_names].to_numpy(dtype=np.float64)),
                         axis=-1,
-                    )
+                    ),
+                    rtol=1e-4,
             ):
                 raise LoggedError(
                     self.log,
@@ -498,7 +499,8 @@ class SampleCollection(BaseCollection):
                     np.sum(
                         np.atleast_2d(self[self.chi2_names].to_numpy(dtype=np.float64)),
                         axis=-1,
-                    )
+                    ),
+                    rtol=1e-4,
             ):
                 raise LoggedError(
                     self.log,
--- a/cobaya/collection.py
+++ b/cobaya/collection.py
@@ -11,7 +11,7 @@ import os
 import functools
 import warnings
 from copy import deepcopy
-from typing import Union, Sequence, Optional, Tuple, List, Sized
+from typing import Union, Sequence, Optional, Tuple, List
 from numbers import Number
 import numpy as np
 import pandas as pd
@@ -217,7 +217,7 @@ class SampleCollection(BaseCollection):
             raise LoggedError(self.log, "'sample_type' must be one of %r.", sample_types)
         self.sample_type = sample_type.lower() if sample_type is not None else sample_type
         self.cache_size = cache_size
-        self.is_batch = False
+        self.is_batch = is_batch
         self._data = None
         self._n = None
         # Create/load the main data frame and the tracking indices
@@ -527,7 +527,7 @@ class SampleCollection(BaseCollection):
         if weights is None:
             weights = [self[OutPar.weight].to_numpy(dtype=np.float64)]
         else:
-            if not isinstance(weights[0], Sized):
+            if not hasattr(weights[0], "__len__"):
                 weights = [weights]
             weights = [np.array(ws) for ws in weights]
             if length is None:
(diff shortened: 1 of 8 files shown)


cmbant:
Looks OK, I guess whatever you do it will be a bit messy

JesusTorrado:
> Looks OK, I guess whatever you do it will be a bit messy

Yeah, no clean solution. Finally added `with_batch` in the minimal number of places where spurious relative weights between chains can be introduced: when detempering and reweighting. I have added a test for this features in `test_post_prior`, on top of the equivalent getdist one already there.

I have also tried to solve #306

Waiting for your review/go-ahead.

cobaya/likelihood.py
```diff
@@ -58,8 +58,7 @@ def current_logp(self) -> float:
         :return:  log likelihood from the current state as a scalar
         """
         value = self.current_state["logp"]
-        # Unfortunately, numpy arrays are not derived from Sequence, so the test is ugly
-        if hasattr(value, "__len__"):
+        if isinstance(value, Sized):
             value = value[0]
```
cmbant, line 62: Sized does not imply indexable, Sequence or MutableSequence?

cobaya/collection.py
```diff
@@ -202,13 +209,15 @@ class SampleCollection(BaseCollection):
 
     def __init__(self, model, output=None, cache_size=_default_cache_size, name=None,
                  extension=None, file_name=None, resuming=False, load=False,
-                 temperature=None, onload_skip=0, onload_thin=1, sample_type=None):
+                 temperature=None, onload_skip=0, onload_thin=1, sample_type=None,
+                 is_batch=False):
```
cmbant, line 213: is_batch is not used?

cobaya/likelihood.py
cmbant: Note the Sized form is also nearly ten times slower than hasattr, which I think may be why we were using that

tests/test_post.py
```diff
@@ -105,8 +132,10 @@ def test_post_prior(tmpdir, temperature):
             new_mean, new_cov = mpi.share()
         # Noisier with higher temperature
         atol, rtol = (1e-8, 1e-5) if temperature == 1 else (5e-4, 1e-3)
```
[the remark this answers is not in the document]
JesusTorrado: Well spotted. This was there since a while ago, recent fixes make it unnecessary. Removing. (If anything, if detempering loaded chains we lose some precision due to the ".8g" txt formatting.)

cobaya/likelihood.py
JesusTorrado: Sized simply implies that it has length (as opposed to Sequence, which is not compatible with numpy arrays). Reverting where applicable. I though you preferred stronger type checks and less duck-typing, but fine to me.

cobaya/collection.py
JesusTorrado: Oops!

JesusTorrado:
OK to merge if tests pass?

Should we merge #319 and release 3.3.3? or better 3.4 since the they were a couple of major fixes?

tests/test_post.py
```diff
@@ -84,29 +106,35 @@ def test_post_prior(tmpdir, temperature):
...
                 info_post["output"] + _post_ + info_post["post"]["suffix"],
-                to_getdist=True
+                to_getdist=True,
             )
             new_mean = mcsamples.mean(["a", "b"])
             new_cov = mcsamples.getCovMat().matrix
             mpi.share((new_mean, new_cov))
         else:
             new_mean, new_cov = mpi.share()
         # Noisier with higher temperature
```
cmbant, line 133: del comment

cmbant:
Sounds good to me, 3.4 could make sense.

cmbant:
Which changes addressed #306?

JesusTorrado:
> Which changes addressed #306?

https://github.com/CobayaSampler/cobaya/pull/322/commits/3957c79d237d1293b908a5d93a58f3576c145ec8

Raises rtol from default 1e-5 to 1e-4. Could not verify, bc I have failed to reproduce the issue myself (reload with custom priors, at different temperatures).

cmbant:
Thanks. I guess it depends on the order of magnitude of the various terms.

PR #326, opened 2023-10-16
New grid scripts and fix cobaya-grid-create

cmbant:
See [doc](https://cobaya.readthedocs.io/en/devel/grids.html) and [sample input script](https://github.com/CobayaSampler/cobaya/blob/grid_update/tests/simple_grid.py)


cmbant:
@JesusTorrado test_cosmo_grid_2.yaml gives "'Varying Neff and total mass of 3 degenerate nu's'with theory code 'classy'.
This does not mean that you cannot use this model with that theory code; just that we have not implemented this yet." - can we just delete this file?

cmbant:
@JesusTorrado, if you have any thoughts on grids now might be the time to comment. I've made a first-pass attempt at supporting python grid settings files, sort-of hybrid between your dict structure and what I had in cosmomc.  My current toy example is (which could be used for fast tests) is
https://github.com/CobayaSampler/cobaya/blob/grid_update/tests/simple_grid.py
Let me know if you have any thoughts/objections about how to best dto this.

I did have to revert some of your changes, e.g. JobItem now stores lists of parameter names and data names as before, rather than just combined name tags - this was needed to make all the job script filters (by param name, or datatag name) work, e.g. when use cobaya-grid-tables etc. (these newly imported scripts otherwise seem to all work fine with cobaya grids)

cmbant:
@JesusTorrado OK to merge this? Most of the changes are orthogonal to the main code (and the grid code currently in master doesn't work)

PR #327, opened 2023-10-18
Add the TDCOSMO likelihood to likelihood_external.rst

nataliehogg:
Adds the hierarchical strong lensing time delay likelihood of TDCOSMO IV ([Birrer et al. 2020](https://arxiv.org/abs/2007.02941)) to the docs.


cmbant:
Do you need to re-instantiate LensSampleLikelihood at each likelihood call?

Issue #337, opened 2023-11-22
`WantTensors: true` in `extra_args` results in a segfault (when also using `external_primordial_pk: true`


lukashergt:
# Issue

I would like to use an external primordial power spectrum with tensor modes.

In `cobaya/theories/camb/camb.py` line 336 it says:
```python
# must set WantTensors manually if using external_primordial_pk
```

However, when I supply `WantTensors: true` to `extra_args` I end up getting a segfault.

# Example

Here is a hopefully reproducible example yaml file:
```yaml
theory:
  theory_primordial_Pk.FeaturePrimordialPk:
    python_path: "."
    k_pivot: 0.05
    n_samples_wavelength: 20
  camb:
    extra_args:
      lens_potential_accuracy: 1
      num_massive_neutrinos: 1
      nnu: 3.044
      theta_H0_range:
      - 20
      - 100
      WantTensors: true
    external_primordial_pk: true
likelihood:
  planck_2018_lowl.TT: null
params:
  amplitude: 0.08
  wavelength: 0.008
  centre: 0.2
  logwidth: 0.1
  # Baseline cosmological parameters
  As: 2.2e-9
  ns: 0.96
  r:
    prior: {min: 0, max: 3}
    ref: {dist: norm, loc: 0, scale: 0.03}
    proposal: 0.03
    latex: r_{0.05}
  H0: 70.0
  ombh2: 0.0224
  omch2: 0.12
  mnu: 0.06
  tau: 0.055
sampler:
  mcmc:
    proposal_scale: 1.9
    oversample_power: 0.4
```

WITHOUT the `WantTensors: true` line, this example runs without complaining (but then the tensor modes are not correctly sampled...).
WITH the `WantTensors: true` line, this example segfaults (see error below).

The `theory_primordial_Pk.FeaturePrimordialPk` I have taken from [here](https://github.com/CobayaSampler/paper_cosmo_demo/blob/main/theory_primordial_Pk.py), modified only by adding `r` everywhere after `ns` and by adding the lines
```python
        state['primordial_tensor_pk'] = {'k': ks, 'Pk': Pks, 'log_regular': False}
```
and
```python
    def get_primordial_tensor_pk(self):
        return self.current_state['primordial_tensor_pk']
```

(Not meant to be physically sensible, just meant to be able to reproduce the error...)

# Error message

This is the error message that I'm getting:
```
[output] Output to be read-from/written-into folder '.', with prefix 'test'
[...]

PR #339, opened 2023-11-22
fix for WantTensors in extra_args and external primordial

cmbant:


lukashergt (approved):
Thanks, Antony, this fixes #337.

PR #343, opened 2023-12-06
[WIP] fixes for parallel calls inside Jupyter

JesusTorrado:
For now just a workaround, that fixes issues with pickling in ipyparallel.

One of the problems can be fixed by not having `__getstate__` of `HasLogger` return a copy, but the original dict minus the logger. This is indeed what a `__getstate__` is supposed to do, but I still need to test this change.

The other fix removes the fast prior pdfs, because the way the method is assigned maybe does not propagate correctly across copies. It can probably be superseded by creating a custom scipy prior object inheriting from the scipy one and having the new logpdf as a first-class method.

EDIT: Copy fixed cherrypicked and added to https://github.com/CobayaSampler/cobaya/pull/349. Only fast-pdf missing.

### DIFF
--- a/cobaya/collection.py
+++ b/cobaya/collection.py
@@ -1194,7 +1194,7 @@ class SampleCollection(BaseCollection):
     # Make it picklable -- formatters are deleted
     # (they will be generated next time txt is dumped)
     def __getstate__(self):
-        attributes = super().__getstate__().copy()
+        attributes = super().__getstate__()
         for attr in ['_numpy_fmts', '_header_formatter']:
             try:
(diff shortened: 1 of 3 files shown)


JesusTorrado:
Hi @ggalloni,

It's a recursion error produced by pickling while in the IPP cluster. You can try if the one-line change to `log.py` solves if in your case.

cmbant:
@ggalloni could you check if this reworking of fast-pdf works? https://github.com/CobayaSampler/cobaya/pull/373

PR #344, opened 2023-12-30
Fixing requisites for bao generic with different kinds of observables

Kushallodha:
Fixing computation of requisites for generic BAO when supplied with measurements_file that contains different kinds of observables at several redshifts. To validate the logic, I tested it on the dummy file in the attachment. The older version incorrectly returns `{'angular_diameter_distance': {'z': array([2.34])}, 'Hubble': {'z': array([2.34])}, 'rdrag': None, 'fsigma8': {'z': array([2.34])}} `.

|    z |   value | observable   |
|-----:|--------:|:-------------|
| 0.29 |     1   | DV_over_rs   |
| 1.31 |     1   | DV_over_rs   |
| 1.31 |     1.5 | F_AP         |
| 0.5  |     1.8 | DV_over_rs   |
| 0.5  |     1.2 | F_AP         |
| 0.7  |     1.1 | DV_over_rs   |
| 0.7  |     1.6 | F_AP         |
| 0.91 |     1.5 | DV_over_rs   |
| 0.91 |     1   | F_AP         |
| 1.49 |     1.5 | DV_over_rs   |
| 2.34 |     2   | DM_over_rs   |
| 2.34 |     5   | DH_over_rs   |
| 2.34 |     2   | f_sigma8     |

[bao_mean_check.txt](https://github.com/CobayaSampler/cobaya/files/13798734/bao_mean_check.txt)

### DIFF
--- a/cobaya/likelihoods/base_classes/bao.py
+++ b/cobaya/likelihoods/base_classes/bao.py
@@ -156,7 +156,7 @@ class BAO(InstallableLikelihood):
     type = "BAO"
 
     install_options = {"github_repository": "CobayaSampler/bao_data",
-                       "github_release": "v2.2"}
+                       "github_release": "v2.3"}
 
     prob_dist_bounds: Optional[Sequence[float]]
     measurements_file: Optional[str] = None
--- a/cobaya/likelihoods/base_classes/bao.py
+++ b/cobaya/likelihoods/base_classes/bao.py
@@ -396,15 +396,15 @@ class BAO(InstallableLikelihood):
                 obs_used_not_implemented, list(theory_reqs))
         requisites = {}
         if self.has_type:
-            for observable in theory_reqs:
-                for req,req_values in theory_reqs[observable].items():
+            for observable in self.data["observable"].unique():
+                for req, req_values in theory_reqs[observable].items():
                     if req not in requisites.keys():
                         requisites[req] = req_values
                     else:
-                        if isinstance(req_values,dict):
-                            for k,v in req_values.items():
+                        if isinstance(req_values, dict):
+                            for k, v in req_values.items():
                                 if v is not None:
-                                    requisites[req][k] = np.unique(np.concatenate((requisites[req][k],v)))
+                                    requisites[req][k] = np.unique(np.concatenate((requisites[req][k], v)))
         return requisites
 
     def theory_fun(self, z, observable):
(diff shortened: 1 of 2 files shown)


cmbant:
Thanks, the tests are failing, but probably just a trivial whitespace issue.

cmbant:
Shouldn't it be something like this?
```
        if self.has_type:
            for observable in self.data["observable"].unique():
                for req, req_values in theory_reqs[observable].items():
                    if req not in requisites:
                        requisites[req] = req_values
                    else:
                        for k, v in req_values.items():
                            if v is not None:
                                requisites[req][k] = np.unique(
                                    np.concatenate((requisites[req][k], v)))
```

Kushallodha:
Yes, thank you @cmbant for pointing this out. It needs to be `for observable in self.data["observable"].unique()` but we still need to check (`if isinstance(req_values, dict)`) req_values is dict before for loop.  Let me know if I missed a better/simplified way to do the same. 
``` 
        if self.has_type:
            for observable in self.data["observable"].unique():
                for req, req_values in theory_reqs[observable].items():
                    if req not in requisites.keys():
                        requisites[req] = req_values
                    else:
                        if isinstance(req_values, dict):
                            for k, v in req_values.items():
                                if v is not None:
                                    requisites[req][k] = np.unique(np.concatenate((requisites[req][k], v)))

```

For the time being, all the checks have passed. Please let me know if you prefer all commit history squashed into one. If not, then I will submit the pull request.

cmbant:
Thanks, why do you need to check it is dict - aren't they all dict? Don't worry about squash, we do that when merging.

Kushallodha:
I think rdrag ( `{"rdrag": None}`) is not the dict.

cmbant:
None is caught by the "if v is not None", the whole of {"rdrag":None} is a dict insance?

cmbant:
(if you have a simple test, it would be good to add it to the unit tests along with the other bao tests)

Kushallodha:
In case of `{"rdrag":None}`, req_values is None and for loop crashes before if condition as None can't be iterated over with .items().

I've created a basic test using mock data for mixed observables, similar to 'test_generic_camb`, which requires a mock mean and data. Please let me know if this is okay or if you have something different in mind.

cmbant:
Thanks, probably just need to update BAO's github_release": to "v2.3" for new data tests to pass.

cmbant:
Yeah, weird - I'm sure it's nothing to do with this PR.

cmbant:
I think it was just a temporary github issue, seems to pass now. Thanks for the PR!

Issue #346, opened 2024-02-07
"The sum of logpriors in the sample is not consistent." when resuming chains


gerrfarr:
I am getting recurring errors stating that "The sum of logpriors in the sample is not consistent." I tracked the issue to the check here
https://github.com/CobayaSampler/cobaya/blob/a3cb9bfb2c63fb5ed1315e9f0f5299cf72e6c8cc/cobaya/collection.py#L467

I can see that the relative tolerance was already increased slightly recently. Would it be a bad idea to also increase the absolute tolerance a bit? It appears that the issue arises when the logprior is close to zero.

cmbant:
Seems like https://github.com/CobayaSampler/cobaya/issues/306 is not fixed. If you have a simple reproducing example that may help Jesus to figure it out (introduced in https://github.com/CobayaSampler/cobaya/pull/276)

gerrfarr:
I don't have one on hand, but I could possibly construct something. I figured out that I can alleviate the problem in my case by simply adjusting the absolute tolerance up to 1e-7 (from the default of 1e-8). I suspect it's one of my custom priors which may have fluctuations which are bit larger than usual

JesusTorrado:
Hi @gerrfarr 

You can try https://github.com/CobayaSampler/cobaya/pull/349 if you have time.

PR #349, opened 2024-02-14
collections: extra tolerance for checks, with warning

JesusTorrado:
Changes to a 2-level tolerance test: the original one, plus a more permissive one with a warning if it fails.


cmbant (approved):
Looks good to me (though not getting to root of why test fails - abs/rel tol?)

JesusTorrado:
Hard to know without a minimal non-working example. and non was provided in https://github.com/CobayaSampler/cobaya/pull/346

PR #377, opened 2024-08-21
GitHub Action live and deploy

cmbant:
For pypi to work, I think just need enable the GitHub connector in Cobaya pypi.org's settings. Presumably similar for coverage, but that needs uncommenting and testing.


cmbant:
@JesusTorrado would be helpful to have this merged soon, so PR tests work consistently again; could you try updating the pypi settings (or add me/make part of shared cobaya org)

PR #385, opened 2024-10-08
Add description for Hillipop/Lollipop likelihoods

mtristram:
Add description for Planck Hillipop+Lollipop NPIPE likelihood in input database.


cmbant:
Thanks for adding! Please put it below the CamSpec NPIPE one so the first option is the same as before (used for some examples).

mtristram:
I switched the order.

Issue #398, opened 2025-01-31
Package install logic


cmbant:
Check lines around here, seems to access string like dictionary if package_install = 'pip' (string):

https://github.com/CobayaSampler/cobaya/blob/5c605d39ab7e0a13f39312b123ffd5fc3d6a2aba/cobaya/install.py#L193

Think these lines were changed las by @JesusTorrado

PR #400, opened 2025-01-31
Proposed fix to issue 398

mgerbino:
This is a proposed fix to #398 with offending lines modified in install.py
Although this works, there might be a smarter and easier way to address the issue.


cmbant:
Thanks, looks sensible (@JesusTorrado can you comment on what your last change was for and whether it does what you intended?)

JesusTorrado:
Hi both,

I will check it as soon as I am able, hopefully by the end of the week.

cmbant:
The test fail is one trivial whitespace, plus a classy error (which is also happening on other tests and nothing to do with Cobaya, though maybe we should re-tag it if they now have a stable version that works with numpy 2)

PR #406, opened 2025-03-11
added pre-commit, pyproject-flake8

ewanchamberlain:
Added pyproject-flake8 with config from docs, pre-commit to run pyproject-flake8 on /cobaya on commit.

Note that line length has not been enforced as there are currently files with lines > 90 characters.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+repos:
+  - repo: https://github.com/csachs/pyproject-flake8
+    rev: v7.0.0
+    hooks:
+      - id: pyproject-flake8
+        files: ^cobaya/.*\.py$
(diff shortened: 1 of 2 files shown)


cmbant:
Thanks, seems like a good idea. Perhaps we should switch to ruff
https://github.com/astral-sh/ruff-pre-commit

ewanchamberlain:
I think that would be a good idea. Here's the Ruff config I'm using in another project. I tried to replicate Cobaya's formatting as much as possible (sadly Ruff didn't have at least one rule available)
https://gist.github.com/ewanchamberlain/397f28277e8a718c91f3d5fb0626371e

cmbant:
I think E704 (multiple on one line is auto-fixable by the ruff formatter, which is why not included) and can use [tool.ruff] entry

src = ["cobaya"]

to include e.g. only source under cobaya (rather than long exclude list)

PR #407, opened 2025-03-19
Adding ACT DR6 Multi-Frequency Likelihood

xgarrido:


cmbant:
Thanks! Do you want to add lite as well?

cmbant:
(not sure why reporting all tests passed, but some checks not completed)

HTJense:
I just came over to make a PR for this myself, but if ACT-lite could be added (https://github.com/ACTCollaboration/DR6-ACT-lite) in this PR that would be great!

I'll see if I can add it to pypi one of these days, but in the meantime people can install it from github.

PR #433, opened 2025-07-09
Add simple Gaussian likelihood with normalized/unnormalized options

### LINKED ISSUE #416: gaussian_mixture norm
Maybe should add option to return chi2, negative value keeps confusing people.

cmbant:
This PR adds a new simple Gaussian likelihood class that addresses issue #416.

## Features

- **Simple Gaussian likelihood** following the gaussian_mixture pattern but much simpler (no mixture)
- **Normalized option** (default True): Includes full normalization constant
- **Unnormalized option**: Only chi-squared term without normalization constant
- **Efficient computation**: Uses precomputed inverse covariance and `chi_squared` function from cobaya.functions
- **Numerical stability**: Uses `numpy.linalg.slogdet` for log determinant calculation
- **Flexible inputs**: Supports scalar, sequence, or ndarray for mean and covariance
- **1D scalar support**: Automatically converts scalars to arrays with `atleast_1d`

## Implementation Details

- Located in `cobaya/likelihoods/gaussian/`
- Proper type annotations: `mean: float | Sequence | np.ndarray`
- No wait/delay parameters (removed unnecessary functionality)
- Supports both explicit `input_params` and `input_params_prefix` approaches

## Testing

- Comprehensive tests in `tests/test_gaussian.py`
- Concise design using loops over normalized/unnormalized cases
- Tests against `scipy.stats.multivariate_normal.logpdf`
- Validates chi-squared calculations for unnormalized case
- Tests both list and numpy array inputs
- Error handling for dimension mismatches and singular matrices

## Addresses Issue #416

As noted in the issue, for a simple Gaussian likelihood, the unnormalized option is well-defined and provides the chi-squared-like piece without the normalization constant. This is useful for cases where only the relative likelihood matters or when the normalization is handled elsewhere.

Fixes #416

---

### DIFF
--- a/cobaya/likelihoods/gaussian/__init__.py
+++ b/cobaya/likelihoods/gaussian/__init__.py
@@ -1,0 +1,1 @@
+from .gaussian import Gaussian as Gaussian
--- a/cobaya/likelihoods/gaussian/gaussian.py
+++ b/cobaya/likelihoods/gaussian/gaussian.py
@@ -1,0 +1,112 @@
+"""
+.. module:: likelihoods.gaussian
+
+:Synopsis: Simple Gaussian likelihood
+:Author: Antony Lewis
+
+"""
+
+import numpy as np
+
+from cobaya.functions import chi_squared
+from cobaya.likelihood import Likelihood
+from cobaya.log import LoggedError
+from cobaya.typing import Sequence
+
+
+class Gaussian(Likelihood):
+    """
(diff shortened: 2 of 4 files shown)


JesusTorrado:
OK, but I would remove most of the tests; not that they are not "correct" but we do definitely not need such granularity, as long as we have tested the code enough ourselves.

In the future, I don't see why not merging both likelihoods.

Issue #438, opened 2025-08-01
Periodic parameters


cmbant:
Allow periodic parameters, e.g.

` {"theta": {"min":0, "max": np.pi, "periodic": True}}
`   

Could be used by sampler, but anyway passed to getdist (cf https://github.com/cmbant/getdist/issues)

PR #440, opened 2025-08-07
[WIP] Periodic parameters

JesusTorrado:
Addresses (#438).

This is just an initial implementation, taking care of the prior part with minimal changes.

It does not modify the original point, e.g. if the period is `[0, 2pi]`, then `3pi` is treated correctly in terms of the prior, but still stored as `3pi`. This could be fixed if we could trust that that the values passed to `Prior.logps_internal` are passed by reference, so that a modification inside the method modifies the actual passed value (otherwise we need to make a copy, as in this initial implementation).

We could also make the code cleaner if we are OK with allowing `periodic: True|False` to pass through `tools.get_scipy_1d_pdf`.

@cmbant thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Added standardized formatting with ruff (removing flake8)
 - Added pre-commit hooks for automated code quality checks (trailing whitespace, end-of-file-fixer, pyupgrade, ruff)
 - Documentation can now be built as markdown for LLM context. Flat context file provided.
+- Periodic sampled parameters implemented (for now fully supported by MCMC only), as well as tagging of derived parameters as periodic for analysis with GetDist.
 
 ## 3.5.7 - 2025-03-31
 
--- a/cobaya/input.py
+++ b/cobaya/input.py
@@ -554,7 +554,7 @@ def merge_params_info(params_infos, default_derived=True):
             current_info[p].update(deepcopy(new_info_p))
             # Account for incompatibilities: "prior" and ("value" or "derived"+bounds)
             incompatibilities = {
-                "prior": ["value", "derived", "min", "max"],
+                "prior": ["value", "derived", "min", "max", "periodic"],
                 "value": ["prior", "ref", "proposal"],
                 "derived": ["prior", "drop", "ref", "proposal"],
             }
--- a/cobaya/model.py
+++ b/cobaya/model.py
@@ -345,6 +345,7 @@ class Model(HasLogger):
         params_values: dict[str, float] | Sequence[float],
         as_dict: bool = False,
         make_finite: bool = False,
+        ignore_periodic: bool = False,
     ) -> np.ndarray | dict[str, float]:
         """
         Takes an array or dictionary of sampled parameter values.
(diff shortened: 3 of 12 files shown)


cmbant:
I was looking at this from the getdist perspective, and it is a bit tricky in general (e.g. the bandwidth selection uses DCT which assumes even, not periodic, boundaries at that bandwidth selection step, so requires significant changes).
I was assuming the parameter samples would be within the stated bounds, so only boundary conditions are changing.

Cobaya could do nothing internally except pass the periodic tag to getdist, but allow the sampler classes to correctly account for it as and when implemented. This would be correct but probably less than optimally efficient.

JesusTorrado:
> I was looking at this from the getdist perspective, and it is a bit tricky in general (e.g. the bandwidth selection uses DCT which assumes even, not periodic, boundaries at that bandwidth selection step, so requires significant changes). I was assuming the parameter samples would be within the stated bounds, so only boundary conditions are changing.

I am not understanding what you mean in that case, sorry. Could you please show me a simple example?

This implementation is anyway useful e.g. for resolvable sources in GW.

> Cobaya could do nothing internally except pass the periodic tag to getdist, but allow the sampler classes to correctly account for it as and when implemented. This would be correct but probably less than optimally efficient.

This is what I have tried to implement in the last 2 commits. `mcmc` checks periodicity now when proposing, so that all points proposed fall within prior definition range. For this to be efficient, that check needs *not* to be performed at log-prior computation. But on the other hand we need the prior itself to support periodic parameters, in case we wanted to call it by itself (not from a sampler). The current solution is to have an `ignore_periodic=True` argument used to skip the test/reduction when calling from a sampler, but ``False`` by default for cases where it's called by hand. It's documented in the relevant docstrings.

For this implementation to be finished in this simple use case, the missing bits are documentation, and enabling it for other samplers (possibly creating a test in the ``__init__`` method of the ``Sampler`` class that would fail for samplers that do not support periodicity).

Do you see any issue for this simple use case?

JesusTorrado:
> > but with boundary condition that the density and derivative should be equal at a and b
> 
> There are no checks for this at the moment, but (except for the derivative case for now) that's easy to implement. Will try.

Done for log-prior value.

cmbant:
This is what I have at the moment for getdist: https://github.com/cmbant/getdist/pull/127

It assumes samples are only within bounds. 
I'm interpreting the prior as being zero outside the quoted rage, but imposing periodic boundary conditions for the range (seems this would be simpler for cobaya that changing the prior - the prior is exactly the same as before, only proposer can (optionally) use periodicity to wrap proposals)

JesusTorrado:
> It assumes samples are only within bounds.

With the current implementation, they are guaranteed to be, no worries on that side. And it's dealt with at the sampler level, not the prior level.

> I'm interpreting the prior as being zero outside the quoted rage, but imposing periodic boundary conditions for the range (seems this would be simpler for cobaya that changing the prior - the prior is exactly the same as before, only proposer can (optionally) use periodicity to wrap proposals)

I guess that's fine for GetDist, as a code where the prior is not a fundamental quantity but an auxiliary function to calculate statistics. But Cobaya is a Bayesian inference code: if `phi in [0, 2pi]` is periodic, then `log prior(3pi)` should return the same as `log prior(pi)`. Otherwise we are just having second-guessing what the prior does.

This is treated exactly the way you say. The modifications in the prior do not play any role during the sampler calls. The rest is extra to having the prior return the correct value in general.

JesusTorrado:
`periodic` added as property for derived parameters.

Still left to do:
- Think about what to do about PolyChord (no support for periodic params, prob. treat as hard boundary and raise warning) and the minimizers.
- Look at tests
- Documentation

JesusTorrado:
> I added getdist support for periodic attribute of prior (as well as parameter) [cmbant/getdist@05f873c](https://github.com/cmbant/getdist/commit/05f873c83c0290c2419989dc00b53ac878a6b365)

Great, thanks!

Now an error is also raised for periodic derived parameters if either `min` or `max` are undefined. But Cobaya does not do anything else at the moment about them.

> I guess parameter-level attribute needs to be allowed for lambda parameters too.

If such as paramters is tracked as derived, it should already be supported in the current implementation. If it is not, then it doesn't show up in GetDist, so the only point of allowing for such value would be to put a reduction-to-base range somewhere, which for a lambda-defined parameter I would say it should be taken care of inside the lambda definition. Or am I missing some other possible use case here?

PR #442, opened 2025-08-13
Linting cleanup

JesusTorrado:
- Removed old linting directives
- Removed bare excepts as a generally skipped error, and added `Exception` (to avoid catching e.g. `KeyboardInterrupt` or `SystemExit`).
- Consistency of development/testing requirements
- ... [WIP]

### DIFF
--- a/DEVEL.rst
+++ b/DEVEL.rst
@@ -20,11 +20,11 @@ Development flow for contributors
 
 1. Fork and clone the repo from github.
 2. From its folder, install in editable mode: ``pip install -e .``
-3. Install development dependencies: ``pip install ruff pre-commit pytest pytest-xdist pytest-cov flaky dill``
+3. Install development dependencies: ``pip install -e .[devel]``
 4. Set up pre-commit hooks: ``pre-commit install``
 5. Modify stuff.
 6. Test with pytest
-7. Make a pull requests and check (about about 15 minutes) if the tests have passed.
+7. Make a pull requests and check (about 15 minutes) if the tests have passed.
 8. Iterate until tests pass, then wait for or request feedback/merge
 
 Contributors must agree to the license (see ``LICENCE.txt`` in the root folder).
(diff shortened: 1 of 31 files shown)


JesusTorrado:
There is a test failing because of an inconsistency: `ruff` is passed through the `tests` folder when it shouldn't (it's not part of the pre-commit hook).

JesusTorrado:
In the context of this PR, I would really, really like to restore the standardised license identifier. This is specially important now that there are miriads of bots scrapping code that may read that field and take it into account, but will possibly ignore it if it's not standard.

It could be done, restoring the original text in the licence file, if

- you (@cmbant) agree to the re-licensing of `mcmc.py`, `proposal.py`, `sn.py` and the `grid_tools` files to L-GPL (there should not be incompatibilities with the CosmoMC license, since there should be really significant changes in all of these cases).
- we put the copyright for the logos somewhere else.
- we put the request for submitting to arxiv somewhere else.

Would that be acceptable?

JesusTorrado:
The last commit seems to work well, including fully standard (SPDX) license identifiers, and keeping out arxiv exception on top of the license file.

The copyright notice for the logos is now under the logos in the readme.

Any objection, or anything related that could be added?

JesusTorrado:
> If we’re including a licence.txt with additional restrictions, then is “LGPL-3.0-or-later AND LicenseRef-GFDL-1.3-or-later” actually valid, as I’d have thought that should imply we have a standard unmodified LGPL licence.txt?

Not sure, but it does not fail. I can still put it somewhere else if you prefer.

> I would guess bots are bimodal – they either totally ignore all licences anyway, or they are a suable (or ethical) entity and then stick very carefully to them.

Maybe, but in any case, I think there is a lot of value in having fully-SPDX-compliant specification. It avoids cases of "well, I tried by my parser didn't get it right because it was non-standard".

cmbant:
Any reason not to squash merge? history would be cleaner and easier to see changes in one go

JesusTorrado:
No reason on my side. I do agree with you. Simply that I wanted to rebase and Github does not give an option to rebase + squash + merge, but you need to choose between rebase or squash. Please let me know if I have missed or misunderstood it and I will use it in general in the future.

PR #443, opened 2025-08-13
Alternative periodic parameter syntax

cmbant:
This PR refactors periodic parameter handling to simplify the API and make the syntax more intuitive.

## Changes

- **Move periodic flag to parameter level**: `periodic: True` is now specified at the parameter level, not inside the prior definition
- **Remove ignore_periodic API**: Eliminated the `ignore_periodic` parameter from all prior/posterior methods (`logpriors`, `logprior`, `logposterior`, `logpost`)
- **Simplify internal structure**: Replace `_is_periodic` array and `is_any_periodic` flag with a compact `_periodic_bounds` list of tuples
- **Update documentation**: Reflect the new parameter-level syntax and clarify that only MCMC currently implements parameter wrapping
- **Remove incompatibility**: `periodic` is no longer incompatible with `prior` in `merge_params_info`

## New Syntax

```yaml
params:
  theta:  # sampled parameter
    prior:
      min: 0
      max: 6.28
    periodic: True  # <- at parameter level
    
  phi:  # derived parameter  
    min: 0
    max: 3.14
    periodic: True  # <- at parameter level
```

## Benefits

- Cleaner, more intuitive syntax
- Simplified API without `ignore_periodic` complexity
- More efficient internal representation
- Better separation of concerns (periodic is a parameter property, not a prior property)

## Backward Compatibility

- The prior itself remains a plain range prior
- `reduce_periodic` method is still available and works as before
- Samplers continue to handle periodic reduction as needed

Follows up on PR #440 which was merged.

---

### DIFF
--- a/cobaya/input.py
+++ b/cobaya/input.py
@@ -554,7 +554,7 @@ def merge_params_info(params_infos, default_derived=True):
             current_info[p].update(deepcopy(new_info_p))
             # Account for incompatibilities: "prior" and ("value" or "derived"+bounds)
             incompatibilities = {
-                "prior": ["value", "derived", "min", "max", "periodic"],
+                "prior": ["value", "derived", "min", "max"],
                 "value": ["prior", "ref", "proposal"],
                 "derived": ["prior", "drop", "ref", "proposal"],
             }
--- a/cobaya/model.py
+++ b/cobaya/model.py
@@ -345,7 +345,6 @@ class Model(HasLogger):
         params_values: dict[str, float] | Sequence[float],
         as_dict: bool = False,
         make_finite: bool = False,
-        ignore_periodic: bool = False,
     ) -> np.ndarray | dict[str, float]:
         """
         Takes an array or dictionary of sampled parameter values.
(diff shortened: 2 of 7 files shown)


cmbant:
I've implemented the optimization suggested by @JesusTorrado to avoid duplicating bounds data.

### Changes

**Before:** `_periodic_bounds` stored tuples of `(index, lower_bound, upper_bound)`
```python
self._periodic_bounds: list[tuple[int, float, float]] = []
...
self._periodic_bounds.append((i, a, b))
```

**After:** `_periodic_bounds` now stores only indices
```python
self._periodic_bounds: list[int] = []
...
self._periodic_bounds.append(i)
```

The bounds are retrieved from `Prior._bounds` when needed in `reduce_periodic()`:
```python
for i in self._periodic_bounds:
    a, b = self._bounds[i]  # Get bounds from main storage
    x[i] = ((x[i] - a) / (b - a)) % 1 * (b - a) + a
```

### Benefits

- ✅ No duplication of bounds data
- ✅ Easier to modify `Prior._bounds` for testing without worrying about sync
- ✅ Cleaner implementation
- ✅ All existing tests pass

---
*Comment posted by [Augment Code](https://www.augmentcode.com/) AI agent*

PR #447, opened 2025-09-01
Add candl, SPT-3G D1 mention to docs

Lbalkenhol:
Add basic mention of the candl CMB likelihood framework to the docs. Also add mention of official SPT-3G D1 likelihoods and link to the relevant websites including the data and tutorials.


cmbant:
The clipy PR is not explicitly using candl, I thought it was just using native python version of clik in clipy (replacing the existing clik wrapper interface provided with cobaya).

Lbalkenhol:
I'm not sure whether you can still load it exactly as before, but I might be wrong. If you have tested that and it works, then it must be fine.

PR #465, opened 2025-11-13
Initial PR to add functionality for new DES-Dovekie release 

bap37:
Good afternoon;

I've added a couple of initial functions to start the process of integrating the new SN likelihood and data for DES-Dovekie. Starting this PR to begin the discussions on what I'm missing/what needs to be added and changed.


bap37:
Added many more necessary code changes ; will need to update the sn-data as well for some processing. the cosmo_tests for SN passes, but I'd like to just fit FLCDM cosmology with SN only to check I loaded everything in appropriately.

PR #477, opened 2026-02-10
add --musthave-params

louisl3grand:
Hello Antony, 
I added an option in the cobaya grid scripts to request --musthave-params, similarly to the --musthave-data.
Cheers, 
Louis


louisl3grand:
I've just found a bug in the jobqueue.py, apparently the arguments for the TextFile() creation and for the .write method were reversed. Thanks claude opus 4.5 for catching that!

cmbant:
Are you sure about TextFile? (imported from getdist, not a standard library).

PR #485, opened 2026-03-20
Ruff test auto-updates and workflow updates

cmbant:


JesusTorrado:
Fixed some of the long lines. It seems there are special cases (in particular regarding string definitions) where `ruff` let long lines through if it cannot find a safe/unambiguous break point.

PR #486, opened 2026-03-24
Adding cwd to path

JesusTorrado:
This modifies the cli scripts to have the same behaviour as calls with `python -m cobaya [command]`, e.g. this now works in both cases:

    likelihood:
      filename.MyLikelihood:


cmbant (commented):
Seems fine, don't forsee any issues.

PR #500, opened 2026-08-11
Add likelihoods with DESI DR2 Lya full-shape results

andreicuceu:
Adding new Lyman-alpha full-shape results from https://arxiv.org/abs/2607.27410. The new likelihoods are:
- `desi_bao_lya_fs`: Lya BAO+FS results (constraints on DV/rd and F_AP)
- `desi_bao_all_lya_fs`: Lya BAO+FS results combined with DESI DR2 BAO measurements from galaxy clustering.


andreicuceu:
Currently, the full DESI DR2 BAO results use `desi_dr2` as shorthand for the `desi_bao_all` likelihood, so we might also want a separate `desi_dr2_fs` for the combination with the full-shape results. I think modifying the current `desi_dr2` to also include full-shape is probably a bad idea, but for future releases, it might be useful to use the shorthands `desi_drx_bao` and `desi_drx_fs` for the joint likelihoods.

